PR #2, opened 2017-05-01
Add public library and buffer utility

natebosch:
- Add `buffer` utility and tests
- Export buffer from package library
- Add travis config

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.0.1-dev
+## 0.0.1
 
 - Add `buffer` utility. Collects events in a `List` until a `trigger` stream
   fires.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,4 @@
+## 0.0.1-dev
+
+- Add `buffer` utility. Collects events in a `List` until a `trigger` stream
+  fires.
--- a/lib/src/buffer.dart
+++ b/lib/src/buffer.dart
@@ -1,0 +1,144 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+import 'dart:async';
+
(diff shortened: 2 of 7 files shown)


natebosch:
Also see earlier discussion at https://github.com/dart-lang/async/pull/29

lib/src/buffer.dart
```diff
@@ -0,0 +1,144 @@
...
+
+    closeController() {
+      var ctl = controller;
+      controller = null;
+      return ctl?.close() ?? new Future.value();
+    }
+
+    emit() {
+      controller.add(currentResults);
+      currentResults = null;
```
jakemac53, line 71: For safety I would just set `waitingForTrigger = true` here, even though it might already be in that state. I think you always want it to be true after this?

CHANGELOG.md
```diff
@@ -0,0 +1,4 @@
+## 0.0.1-dev
```
kevmoo: Don't put `-dev` in the changelog – just in the pubspec.

kevmoo (approved):
LGTM from a structural perspective.

test/buffer_test.dart
```diff
@@ -0,0 +1,144 @@
...
+          values.add(4);
+          await new Future(() {});
+          trigger.add(null);
+          await new Future(() {});
+          expect(emittedValues, [
+            [1, 2],
+            [3, 4]
+          ]);
+        });
+
```
jakemac53, line 126: A couple other suggestions for tests:

- mixing pre-emptive triggers and normal triggers
- multiple triggers in a row in between values

jakemac53 (commented):
Couple quick suggestions for some additional tests, but otherwise lgtm.

lib/src/buffer.dart
natebosch: Good suggestion. Done

test/buffer_test.dart
natebosch: Added 3 new tests around pre-emptive trigger behavior

jakemac53 (approved):

PR #3, opened 2017-05-01
Add `debounce` utilities and prepare for publish

natebosch:
- Add `debounce` and `debounceBuffer` to drop or collect values that
  occur quickly
- Add tests
- Drop `-dev` from pubspec to prepare for publishing very early version

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,3 +11,12 @@ dart_task:
   - test
   - dartfmt
   - dartanalyzer
+matrix:
+  include:
+    - dart: dev
+      dart_task: dartfmt
+    # Only care about being analyzer clean for dev and stable
+    - dart: dev
+      dart_task: dartanalyzer
+    - dart: stable
+      dart_task: dartanalyzer
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,6 @@ dart_task:
 matrix:
   exclude:
     - dart: 1.22.1
-      dart_task:
-        - dartfmt
-        - dartanalyzer
+      dart_task: dartfmt
+    - dart: 1.22.1
+      dart_task: dartanalyzer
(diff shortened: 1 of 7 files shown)


kevmoo:
@natebosch now you've hit a spot where you'll want to skip `1.22.1` release – at least for analyzer and dartfmt

.travis.yml
```diff
@@ -9,5 +9,12 @@ cache:
     - $HOME/.pub-cache
 dart_task:
   - test
-  - dartfmt
```
kevmoo: Do it the other way: exclude only 1.22.1 😄

kevmoo (changes requested):

kevmoo (approved):

matanlurey:
Yay! This closes https://github.com/dart-lang/core/issues/296.

lib/src/debounce.dart
```diff
@@ -0,0 +1,57 @@
...
+      if (shouldClose) {
+        sink.close();
+      }
+      soFar = null;
+      timer = null;
+    });
+    soFar = collect(value, soFar);
+  }, handleDone: (EventSink<T> sink) {
+    if (soFar != null) {
+      shouldClose = true;
```
jakemac53, line 52: Given that you know there will be no more events at this point, it might make sense to just immediately add `soFar` to `sink` and cancel the timer?

natebosch: I could. It would break the assumption that "all events are delayed by at least [duration]". Do you think it is worth the additional model complexity to do so?

jakemac53: Ya its true this could result in multiple events coming faster than you would expect, probably best to leave as is.

jakemac53 (approved):

PR #7, opened 2017-05-05
Implement merge and mergeAll

### LINKED ISSUE #4: Implement `merge` and `mergeAll`
Reference: http://reactivex.io/documentation/operators/merge.html.

### LINKED ISSUE #1747: Implement `merge` and `mergeAll`
Reference: http://reactivex.io/documentation/operators/merge.html.

natebosch:
Closes dart-lang/tools#1747 

merge and mergeAll combine/interleave streams. Handles both broadcast
and single-subscription streams.

Add implementation of containsAllInOrder in test/util - once this is
submitted in package:matcher it can be removed here. Submitting here
since the other package is blocked.

Remove -dev from pubspec for publishing.

### DIFF
--- a/lib/src/merge.dart
+++ b/lib/src/merge.dart
@@ -1,0 +1,71 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+
+/// Emits values from the source stream and [other] in any order as they arrive.
+///
+/// If the source stream is a broadcast stream, the result stream will be as
+/// well, regardless of the type of stream [other] is.
+StreamTransformer<T, T> merge<T>(Stream<T> other) => new _Merge<T>([other]);
+
+/// Emits values from the source stream and all streams in [others] in any order
+/// as they arrive.
+///
+/// If the source stream is a broadcast stream, the result stream will be as
+/// well, regardless of the types of streams in [others].
+StreamTransformer<T, T> mergeAll<T>(List<Stream<T>> others) =>
+    new _Merge<T>(others);
+
+class _Merge<T> implements StreamTransformer<T, T> {
+  final List<Stream<T>> _others;
+
+  _Merge(this._others);
+
+  @override
+  Stream<T> bind(Stream<T> first) {
+    StreamController<T> controller;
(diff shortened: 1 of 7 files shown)


natebosch:
Not sure why we didn't get the travis failure before... Fixed by https://github.com/dart-lang/stream_transform/pull/8

lib/src/merge.dart
```diff
@@ -0,0 +1,71 @@
...
+///
+/// If the source stream is a broadcast stream, the result stream will be as
+/// well, regardless of the type of stream [other] is.
+StreamTransformer<T, T> merge<T>(Stream<T> other) => new _Merge<T>([other]);
+
+/// Emits values from the source stream and all streams in [others] in any order
+/// as they arrive.
+///
+/// If the source stream is a broadcast stream, the result stream will be as
+/// well, regardless of the types of streams in [others].
```
grouma, line 17: What was the reasoning behind this? Why not check the type of others?

test/merge_test.dart
```diff
@@ -0,0 +1,102 @@
...
+          first.stream.transform(merge(second.stream)).listen((_) {});
+      await subscription.cancel();
+      expect(firstCanceled, true);
+      expect(secondCanceled, true);
+    });
+  });
+
+  group('mergeAll', () {
+    test('includes all values', () async {
+      var first = new Stream.fromIterable([1, 2, 3]);
```
grouma, line 41: Given the values in the various iterables, it's not strictly necessary, but you should have a test to insure the correct number of values are contained in the merged stream.

grouma (approved):

lib/src/merge.dart
natebosch: In general I think it *can* be sensible to merge streams of mixed types so I want to support it. We have to pick a behavior and picking it based on the source stream seems the most sensible to me.

I think it's most common for a `StreamTransformer` to preserve the type of the source stream. That's what `new StreamTransformer.fromHandlers()` would do which we use elsewhere in this package. I suppose we could drop this part of the comment altogether and most users will assume it, but I thought it might be worthwhile to make it explicit since there are multiple streams in question here.

test/merge_test.dart
natebosch: I could add a 'does not add extra values' test... I would be surprised if we found a way to introduce that bug but it won't hurt to test :)

PR #11, opened 2017-05-06
Adds Throttle, Audit, and Scan

jonahwilliams:
Throttle: emits immediately, then swallows values until duration elapses

Audit: like throttle, but it emits the most recent value at the end of the duration.

Scan: a fold which yield the intermediate accumulations.

cc @natebosch for review?

### DIFF
--- a/lib/src/audit.dart
+++ b/lib/src/audit.dart
@@ -5,6 +5,9 @@ import 'dart:async';
 
 /// Creates a StreamTransformer which only emits once per [duration], at the
 /// end of the period.
+///
+/// Like `throttle`, except it always emits the most recently received event in
+/// a period.  Alwyas introduces a delay of at most [duration].
 StreamTransformer<T, T> audit<T>(Duration duration) {
   Timer timer;
   bool shouldClose = false;
--- a/lib/src/audit.dart
+++ b/lib/src/audit.dart
@@ -7,7 +7,7 @@ import 'dart:async';
 /// end of the period.
 ///
 /// Like `throttle`, except it always emits the most recently received event in
-/// a period.  Alwyas introduces a delay of at most [duration].
+/// a period.  Always introduces a delay of at most [duration].
 StreamTransformer<T, T> audit<T>(Duration duration) {
   Timer timer;
   bool shouldClose = false;
--- a/lib/src/audit.dart
+++ b/lib/src/audit.dart
@@ -1,0 +1,35 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+import 'dart:async';
+
+/// Creates a StreamTransformer which only emits once per [duration],
+/// at the end of the period.
+///
+StreamTransformer<T, T> audit<T>(Duration duration) => _auditer(duration);
+
+StreamTransformer<T, T> _auditer<T>(Duration duration) {
+  Timer timer;
+  bool shouldClose = false;
+  T recentData;
+
+  return new StreamTransformer.fromHandlers(
+      handleData: (T data, EventSink<T> sink) {
+    recentData = data;
+    if (timer == null) {
+      timer = new Timer(duration, () {
+        sink.add(recentData);
+        timer = null;
+        if (shouldClose) {
+          sink.close();
+        }
+      });
+    }
+  }, handleDone: (EventSink<T> sink) {
+    if (timer != null) {
+      shouldClose = true;
+    } else {
+      sink.close();
+    }
+  });
+}
--- a/lib/src/audit.dart
+++ b/lib/src/audit.dart
@@ -3,12 +3,9 @@
 // BSD-style license that can be found in the LICENSE file.
 import 'dart:async';
 
-/// Creates a StreamTransformer which only emits once per [duration],
-/// at the end of the period.
-///
-StreamTransformer<T, T> audit<T>(Duration duration) => _auditer(duration);
-
-StreamTransformer<T, T> _auditer<T>(Duration duration) {
+/// Creates a StreamTransformer which only emits once per [duration], at the
+/// end of the period.
+StreamTransformer<T, T> audit<T>(Duration duration) {
   Timer timer;
   bool shouldClose = false;
   T recentData;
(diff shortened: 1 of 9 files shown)


lib/src/audit.dart
```diff
@@ -0,0 +1,35 @@
...
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+import 'dart:async';
+
+/// Creates a StreamTransformer which only emits once per [duration],
+/// at the end of the period.
+///
+StreamTransformer<T, T> audit<T>(Duration duration) => _auditer(duration);
+
+StreamTransformer<T, T> _auditer<T>(Duration duration) {
```
natebosch: why do we need a second method for this?

lib/src/throttle.dart
```diff
@@ -0,0 +1,24 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+import 'dart:async';
+
+/// Creates a StreamTransformer which only emits once per [duration],
+/// at the beginning of the period.
```
natebosch: [nit] loks like at least 'at' could fit on the previous line. Can you reflow to 80 characters

lib/src/audit.dart
```diff
@@ -0,0 +1,35 @@
...
+  Timer timer;
+  bool shouldClose = false;
+  T recentData;
+
+  return new StreamTransformer.fromHandlers(
+      handleData: (T data, EventSink<T> sink) {
+    recentData = data;
+    if (timer == null) {
```
natebosch: [nit] how about `timer ??=`

lib/src/scan.dart
```diff
@@ -0,0 +1,26 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+import 'dart:async';
+
+typedef S Func2<T, S>(T item, S accumulation);
```
natebosch: we should avoid public typedefs in this package. Can you inline this in the function signature?

lib/src/scan.dart
```diff
@@ -0,0 +1,26 @@
...
+
+/// Scan is like fold, but instead of producing a single value it yields
+/// each intermediate accumulation.
+StreamTransformer<T, S> scan<T, S>(Func2<T, S> map, S accumulation) =>
+    new _Scan(map, accumulation);
+
+class _Scan<T, S> implements StreamTransformer<T, S> {
```
natebosch: [optional] I think this one can be done without a new class since it can be modeled with `map`

```
StreamTransformer<S, T> scan<S, T>(
        T initialValue, T combine(T previousValue, S element)) =>
    new StreamTransformer<S, T>((stream, cancelOnError) {
      T accumulated = initialValue;
      return stream
          .map((value) => accumulated = combine(accumulated, value))
          .listen(null, cancelOnError: cancelOnError);
    });
````

File: lib/src/scan.dart [hunk 6]
```diff
@@ -0,0 +1,26 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+import 'dart:async';
+
+typedef S Func2<T, S>(T item, S accumulation);
+
+/// Scan is like fold, but instead of producing a single value it yields
+/// each intermediate accumulation.
+StreamTransformer<T, S> scan<T, S>(Func2<T, S> map, S accumulation) =>
```
natebosch: how about `scan(S initialValue, S combine(S previous, T element))` so that it matches the signature of `fold`?

lib/src/audit.dart
```diff
@@ -0,0 +1,35 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+import 'dart:async';
+
+/// Creates a StreamTransformer which only emits once per [duration],
```
natebosch: I'm not sure I understand the use case - when might we choose to use this over `throttle`?

test/throttle_test.dart
```diff
@@ -0,0 +1,80 @@
...
+
+        test('outputs multiple values spaced further than duration', () async {
+          values.add(1);
+          await new Future.delayed(const Duration(milliseconds: 10));
+          values.add(2);
+          await new Future.delayed(const Duration(milliseconds: 10));
+          expect(emittedValues, [1, 2]);
+        });
+
+        test('waits for pending value to close', () async {
```
natebosch: how do we know it's waited for the pending value? We don't have an `expect` before the value has been emitted

natebosch (changes requested):

lib/src/audit.dart
jonahwilliams: A good example might be a text field which auto updates after being modified.  Audit will make more regular updates than debounce, but still always emit the most recent value.

Another way to (mostly) solve the same problem would be to give throttle an optional parameter like `emitLast` which ensures that the last value of the stream is always emitted.

lib/src/audit.dart
jonahwilliams: No reason, I will fix this here and in throttle

lib/src/audit.dart
natebosch: Ah cool. I think it could be helpful to call out the differences from debounce and throttle in the doc comment.

lib/src/scan.dart
jonahwilliams: Done.

Note to self, remember that assignment is an expression

test/throttle_test.dart
jonahwilliams: Actually I think this test doesn't apply to throttle, since it never waits.  Removed.

lib/src/scan.dart
jonahwilliams: Done

lib/src/scan.dart
jonahwilliams: Done

lib/src/throttle.dart
jonahwilliams: Done

lib/src/audit.dart
jonahwilliams: Done

natebosch (approved):
Looks good.

Thanks for the contribution!

Could you also update README and CHANGELOG?

jonahwilliams:
Done

lib/src/audit.dart
```diff
@@ -0,0 +1,33 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+import 'dart:async';
+
+/// Creates a StreamTransformer which only emits once per [duration], at the
+/// end of the period.
+///
+/// Like `throttle`, except it always emits the most recently received event in
+/// a period.  Alwyas introduces a delay of at most [duration].
```
natebosch: One last nit: typo in Always.

jonahwilliams: Woops, Done

PR #13, opened 2017-05-08
Add `asyncWhere`

natebosch:
Relatively few tests since most of the complex stuff is handled by
StreamTransformer.fromHandlers


grouma (approved):
Optionally add a test with an explicit microtask.

natebosch:
Good catch - there were bugs when I didn't go around the event loop

grouma (approved):

PR #17, opened 2017-05-25
Add a package vision doc

natebosch:
These ideas are not immutable - they are strong opinions held weakly.
The intent behind documenting in this way is to help keep high
conceptual integrity while the package is young.


natebosch:
Not 100% sure it makes sense to merge this in - but this seemed like the most useful place to stick these thoughts while I may not be communicating regularly.

@kevmoo @matanlurey - If you think it's worthwhile feel free to take this PR over.

PR #22, opened 2017-06-22
Fix broadcast bug for Audit and clarify behavior

natebosch:
- Use `fromHandlers` to correctly handle broadcast Streams.
- Clarify in the Doc comment the behavior difference from `debounce`
- Add a test which makes explicit the difference from `debounce`, all
  the other tests also pass with `debounce`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,10 @@
 ## 0.0.6
 
-- Bug Fix: `debounce` correctly adds data to all listeners on a broadcast
-  stream.
+- Bug Fix: Some transformers did not correctly add date to all listeners on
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,7 +1,10 @@
 ## 0.0.6
 
-- Bug Fix: `debounce` correctly adds data to all listeners on a broadcast
-  stream.
+- Bug Fix: Some transformers did not correctly add date to all listeners on
```
kevmoo: `date` -> `data`?

natebosch: fixed

kevmoo (approved):

PR #23, opened 2017-06-22
Fix throttle broadcast bug

natebosch:
- Switch to `fromHandlers` utility.
- Add a test which fails with the old implementation.
- Add an extra `await new Future(() {})` in the test for handling the
  Stream done event. The new StreamTransformer is not sync like this old
  one - this behavior is still correct.
- Remove extra `handleDone` which matches the default behavior.

### DIFF
--- a/test/throttle_test.dart
+++ b/test/throttle_test.dart
@@ -14,6 +14,7 @@ void main() {
       bool valuesCanceled;
       bool isDone;
       List errors;
+      Stream transformed;
       StreamSubscription subscription;
 
       void setUpStreams(StreamTransformer transformer) {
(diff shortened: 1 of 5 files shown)


natebosch:
We could drop the extra `await` after addressing dart-lang/tools#1754

test/throttle_test.dart
```diff
@@ -64,8 +65,20 @@ void main() {
           values.add(2);
           await new Future(() {});
           await values.close();
+          await new Future(() {});
```
lrhn: This just wants to wait until the end of the current microtask queue.
You can use either `await null;` or `await new Future.value(null);` to the same effect, which should be cheaper than allocating a closure and calling it later.

grouma (approved):

matanlurey (approved):

jakemac53 (approved):

PR #26, opened 2017-06-23
Call `tap` data callback once per event

natebosch:
This is slightly different from the other bugs. In this case the output
stream correctly forwarded values to all listeners, but the callback was
also called once per listener.

- Add `handleError` to `fromHandlers` and tests.
- Add tests for expected single-call behavior. These fail on the old
  implementation.
- Add test for the values getting to multiple listeners. This passes on
  both the old and new implementation.
- Switch to `fromHandlers`.

### DIFF
--- a/test/from_handlers_test.dart
+++ b/test/from_handlers_test.dart
@@ -136,18 +136,23 @@ void main() {
     group('broadcast stream with multiple listeners', () {
       int dataCallCount;
       int doneCallCount;
+      int errorCallCount;
 
       setUp(() async {
         dataCallCount = 0;
         doneCallCount = 0;
+        errorCallCount = 0;
         setUpForController(
             new StreamController.broadcast(),
             fromHandlers(handleData: (value, sink) {
               dataCallCount++;
+            }, handleError: (error, stackTrace, sink) {
+              errorCallCount++;
+              sink.addError(error, stackTrace);
             }, handleDone: (sink) {
               doneCallCount++;
             }));
-        transformed.listen((_) {});
+        transformed.listen((_) {}, onError: (_, __) {});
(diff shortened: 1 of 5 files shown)


test/from_handlers_test.dart
```diff
@@ -136,18 +136,23 @@ void main() {
...
             fromHandlers(handleData: (value, sink) {
               dataCallCount++;
+            }, handleError: (error, stackTrace, sink) {
+              errorCallCount++;
+              sink.addError(error, stackTrace);
             }, handleDone: (sink) {
               doneCallCount++;
             }));
-        transformed.listen((_) {});
+        transformed.listen((_) {}, onError: (_, __) {});
```
grouma, line 155: I'm not familiar with the double underline syntax. What does it do? Or is this a typo?

grouma (approved):

test/from_handlers_test.dart
natebosch: When there are multiple unused arguments in the signature I use 1 underscore for the first argument, 2 for the next, etc. Luckily this doesn't come up often.

grouma: Nothing is preventing you from using a single underscore for both arguments though eh?

natebosch: I think you get an error saying that `'_'` is already defined.

kevmoo: Well, it'd be like using `a, a` in the arg list. Equally unsupported. 😄

grouma: Ahh. I always thought underscore was special syntax as opposed to convention. Thanks for the explanation.

PR #27, opened 2017-06-27
Use sync forwarders in `fromHandlers`

natebosch:
Towards dart-lang/tools#1754

- Drop the return values from the onCancel callback. For
  single-subscription StreamController instances this is in the path
  before onDone is forwarded so we don't want or need to wait for a
  Future here.
- Drop the now unnecessary `await new Future(() {})` from tests.

### DIFF
--- a/lib/src/from_handlers.dart
+++ b/lib/src/from_handlers.dart
@@ -46,9 +46,9 @@ class _StreamTransformer<S, T> implements StreamTransformer<S, T> {
   Stream<T> bind(Stream<S> values) {
     StreamController<T> controller;
     if (values.isBroadcast) {
-      controller = new StreamController<T>.broadcast();
+      controller = new StreamController<T>.broadcast(sync: true);
     } else {
-      controller = new StreamController<T>();
+      controller = new StreamController<T>(sync: true);
     }
     StreamSubscription<S> subscription;
     controller.onListen = () {
@@ -67,12 +67,9 @@ class _StreamTransformer<S, T> implements StreamTransformer<S, T> {
       controller.onResume = () => subscription?.resume();
     }
     controller.onCancel = () {
-      if (controller.hasListener || subscription == null) {
-        return new Future.value();
-      }
-      var toCancel = subscription;
+      if (controller.hasListener || subscription == null) return;
+      subscription.cancel();
       subscription = null;
-      return toCancel.cancel();
     };
     return controller.stream;
   }
--- a/lib/src/from_handlers.dart
+++ b/lib/src/from_handlers.dart
@@ -44,32 +44,31 @@ class _StreamTransformer<S, T> implements StreamTransformer<S, T> {
 
   @override
   Stream<T> bind(Stream<S> values) {
-    StreamController<T> controller;
-    if (values.isBroadcast) {
-      controller = new StreamController<T>.broadcast(sync: true);
-    } else {
-      controller = new StreamController<T>(sync: true);
-    }
+    var controller = values.isBroadcast
+        ? new StreamController<T>.broadcast(sync: true)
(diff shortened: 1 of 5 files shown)


lib/src/from_handlers.dart
```diff
@@ -67,12 +67,9 @@ class _StreamTransformer<S, T> implements StreamTransformer<S, T> {
       controller.onResume = () => subscription?.resume();
```
lrhn: You shouldn't need the `?.` here. The `subscription` is set to non-null in `onListen` above, and is only set to `null` in the `onCancel` callback. There will not be a call to `onPause` or `onResume` before a call to `onListen` or after a call to `onCancel`.
(You can even wait and only set the `onPause`/`onResume`/`onCancel` callbacks inside the `onListen`, they won't be called before that one returns, and then you can do:
   controller.onPause = subscription.pause;
   controller.onResume = subscription.resume;
without the extra function wrapping.)

lib/src/from_handlers.dart
```diff
@@ -67,12 +67,9 @@ class _StreamTransformer<S, T> implements StreamTransformer<S, T> {
...
     }
     controller.onCancel = () {
-      if (controller.hasListener || subscription == null) {
-        return new Future.value();
-      }
-      var toCancel = subscription;
+      if (controller.hasListener || subscription == null) return;
```
lrhn: The `controller.hasListener` will always be false in the `onCancel` call.
Even a broadcast stream doesn't get an `onCancel` callback for each listener, it only gets one when there is no listener left.

lib/src/from_handlers.dart
```diff
@@ -67,12 +67,9 @@ class _StreamTransformer<S, T> implements StreamTransformer<S, T> {
...
+      subscription.cancel();
```
lrhn: The `subscription.cancel` call *can* return a future – it comes from the `Stream` you took as argument, so you have no idea whether it returns a future or not – and that future can contain an error, so you *should* forward it so the error can be properly handled.

lrhn (commented):
drive-by-comments.

lib/src/from_handlers.dart
natebosch: If I forward the returned Future from `subscription.cancel()` then there is an extra microtask between the source stream finishing it's `.close()` and the result Stream calling it's subscribers `onDone`.

Ideally I'd like `onDone` behavior consistent with the source stream. I can get that with a sync broadcast stream, or by ignoring the Future here - but I don't get it with a single-subscriber stream and returning this Future.

See this in action here: https://dartpad.dartlang.org/35319efd7419556622f76c9cc124b9d7

If this is an unreasonable goal - if the onDone callbacks are not meant to complete within the Future returned by the source controller `.close()` - then I suppose I could add back some extra `await`s in the tests...

natebosch:
Thanks for the pointers!

I'll take another stab at this pending the discussion about the interaction between `onCancel` and `.close()`.

lib/src/from_handlers.dart
natebosch: Done.

lib/src/from_handlers.dart
natebosch: Much nicer. Done.

lib/src/from_handlers.dart
natebosch: I think I found an approach that won't lose this Future. I shouldn't need to cancel the subscription if the `onCancel` is triggered by the stream finishing - so in the case where the values are done I'll skip it, otherwise I'll use that as the return value.

natebosch:
Updated without the misleading semantics - @grouma PTAL

grouma (approved):

PR #29, opened 2017-06-29
Ensure timers are started before awaits in tests

### LINKED ISSUE #25: Flaky tests on Travis
The `debounce` and `audit` tests are flaky on Travis because we rely on a 10ms delay firing after a 5ms delay.

It's possible that #24 would make this more reliable. If not we could also consider bumping the tests to 15ms wait.

Filing for tracking but won't spend time on this until after we try sync controllers.

### LINKED ISSUE #1755: Flaky tests on Travis
The `debounce` and `audit` tests are flaky on Travis because we rely on a 10ms delay firing after a 5ms delay.

It's possible that dart-lang/tools#1754 would make this more reliable. If not we could also consider bumping the tests to 15ms wait.

Filing for tracking but won't spend time on this until after we try sync controllers.

natebosch:
Fixes dart-lang/tools#1755

These tests were relying on a 5ms Timer getting started asynchronously
and firing before a 10ms delay which is not guaranteed. By waiting an
extra cycle through the event loop to ensure that the Timer starts
before the delay we can have a better guarantee of ordering.

### DIFF
--- a/test/throttle_test.dart
+++ b/test/throttle_test.dart
@@ -1,3 +1,7 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
 import 'dart:async';
 
 import 'package:test/test.dart';
--- a/test/throttle_test.dart
+++ b/test/throttle_test.dart
@@ -47,21 +47,21 @@ void main() {
           values.add(1);
           values.add(2);
           await values.close();
-          await new Future.delayed(const Duration(milliseconds: 10));
+          await _waitForTimer(5);
           expect(emittedValues, [1]);
         });
 
         test('outputs multiple values spaced further than duration', () async {
           values.add(1);
-          await new Future.delayed(const Duration(milliseconds: 10));
+          await _waitForTimer(5);
           values.add(2);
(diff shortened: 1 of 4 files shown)


natebosch:
We could also potentially look at using sync StreamControllers in these tests so that the Timer is started synchronously instead - but I'm not sure I'm comfortable running tests with synchronous input streams when that won't be the behavior we see in the real world. Instead I've added an extra loop around the event queue which I think should make the test reliable - waiting on Travis to see.

test/throttle_test.dart
```diff
@@ -81,3 +81,9 @@ void main() {
     });
   }
 }
+
+/// Cycle the event loop to ensure timers are started, then wait for a delay
+/// longer than [milliseconds] to allow for the timer to fire.
+Future _waitForTimer(int milliseconds) =>
+    new Future(() {/* ensure Timer is started*/}).then((_) =>
+        new Future.delayed(new Duration(milliseconds: milliseconds + 1)));
```
grouma: Move this to a testing utility file instead of repeating yourself.

grouma:
As a side note, I have added retry functionality to package:test. Although in this case I don't think it makes sense to use it.

test/throttle_test.dart
natebosch: Done

grouma (approved):

PR #31, opened 2017-07-02
Use sync forwarders in buffer

natebosch:
Towards dart-lang/tools#1754

Fixes a bug where the output would close too soon if there was a pending
trigger.

Fixes some bugs in the test.

Fixes a bug with canceling when values are broadcast but trigger is
single subscription.

Refactor for better readability

- Fix tests to pass the valuesType instead of the triggerType for both
  streams.
- Refactor for more clear parallel structure between handlers for values
  and triggers.
- Inline _collectToList
- Add booleans to track when streams are done rather than overload what
  it means for their subscriptions to be null.
- Only return a Future from onCancel when there were streams open that
  needed to be canceled. This means that some of the extra awaits in
  tests need to be left in to properly forward errors closing stream.
- Pull the trigger canceling tests out of the loops since the behavior
  depends on both stream types.
- Cancel rather than pause trigger subscription if we won't need it
  again.
- Add tests that broadcast stream (from values) can be canceled and
  relistened
- Set up onPause, onResume, and onCancel in onListen

### DIFF
--- a/lib/src/buffer.dart
+++ b/lib/src/buffer.dart
@@ -14,12 +14,6 @@ import 'dart:async';
 /// the output.
 StreamTransformer<T, List<T>> buffer<T>(Stream trigger) => new _Buffer(trigger);
 
-List<T> _collectToList<T>(T element, List<T> soFar) {
-  soFar ??= <T>[];
-  soFar.add(element);
-  return soFar;
-}
-
 /// A StreamTransformer which aggregates values and emits when it sees a value
 /// on [_trigger].
 ///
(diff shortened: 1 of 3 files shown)


natebosch:
ping @grouma

lib/src/buffer.dart
```diff
@@ -36,109 +30,95 @@ class _Buffer<T> implements StreamTransformer<T, List<T>> {
...
-      emit();
-      if (valuesSub == null) {
-        closeController();
-        cancelTrigger();
+      waitingForTrigger = false;
+
+      if (currentResults != null) emit();
+
+      if (isValueDone) {
+        triggerSub.cancel();
```
grouma, line 75: This is guaranteed to be non-null?

grouma (approved):

lib/src/buffer.dart
natebosch: If it was null we wouldn't be getting any events through it, so here yes I think it is guaranteed.

PR #32, opened 2017-07-02
Rewrite concat

natebosch:
Uses sync forwarders for dart-lang/tools#1754

The simple implementation leaves out a lot of edge cases around
single-subscription vs broadcast streams and how they combine. Rewrite
to handle pause, resume, and cancel manually.

- Loop over stream types in tests
- Add test for onDone and canceling and relistening
- Handle mix of stream types by choosign whether to cancel or pause as
  appropriate.

### DIFF
--- a/lib/src/concat.dart
+++ b/lib/src/concat.dart
@@ -8,6 +8,9 @@ import 'dart:async';
 ///
 /// If the initial stream never finishes, the [next] stream will never be
 /// listened to.
+///
+/// If a single-subscription is concatted to the end of a broadcast stream it
+/// may be listened to and never canceled.
 StreamTransformer<T, T> concat<T>(Stream<T> next) => new _Concat<T>(next);
 
 class _Concat<T> implements StreamTransformer<T, T> {
@@ -17,11 +20,66 @@ class _Concat<T> implements StreamTransformer<T, T> {
 
   @override
   Stream<T> bind(Stream<T> first) {
-    var controller = new StreamController<T>();
-    controller
-        .addStream(first)
-        .then((_) => controller.addStream(_next))
-        .then((_) => controller.close());
+    var controller = first.isBroadcast
+        ? new StreamController<T>.broadcast(sync: true)
+        : new StreamController<T>(sync: true);
+
+    StreamSubscription subscription;
+    var currentStream = first;
+    var firstDone = false;
+    var secondDone = false;
+
+    Function currentDoneHandler;
+
+    listen() {
+      subscription = currentStream.listen(controller.add,
+          onError: controller.addError, onDone: () => currentDoneHandler());
+    }
+
+    onSecondDone() {
+      secondDone = true;
+      controller.close();
+    }
+
+    onFirstDone() {
+      firstDone = true;
+      currentStream = _next;
+      currentDoneHandler = onSecondDone;
+      listen();
+    }
+
+    currentDoneHandler = onFirstDone;
+
+    controller.onListen = () {
+      if (firstDone &&
+          !_next.isBroadcast &&
+          subscription != null &&
+          subscription.isPaused) {
+        subscription.resume();
+      }
+      if (subscription != null) return;
+      listen();
+      if (!first.isBroadcast) {
+        controller.onPause = () {
+          if (!firstDone || !_next.isBroadcast) return subscription.pause();
+          subscription.cancel();
+          subscription = null;
+        };
+        controller.onResume = () {
+          if (!firstDone || !_next.isBroadcast) return subscription.resume();
+          listen();
+        };
+      }
+      controller.onCancel = () {
+        if (secondDone) return null;
+        if (!firstDone || _next.isBroadcast) {
+          var toCancel = subscription;
+          subscription = null;
+          return toCancel.cancel();
+        }
+        subscription.pause();
+      };
+    };
     return controller.stream;
   }
 }
--- a/lib/src/concat.dart
+++ b/lib/src/concat.dart
@@ -24,6 +24,10 @@ class _Concat<T> implements StreamTransformer<T, T> {
         ? new StreamController<T>.broadcast(sync: true)
         : new StreamController<T>(sync: true);
 
+    var next = first.isBroadcast && !_next.isBroadcast
+        ? _next.asBroadcastStream()
+        : _next;
+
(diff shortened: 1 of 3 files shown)


lib/src/concat.dart
```diff
@@ -17,11 +20,66 @@ class _Concat<T> implements StreamTransformer<T, T> {
...
+      firstDone = true;
+      currentStream = _next;
+      currentDoneHandler = onSecondDone;
+      listen();
+    }
+
+    currentDoneHandler = onFirstDone;
+
+    controller.onListen = () {
+      if (firstDone &&
```
jakemac53: I am a bit confused by a lot of the logic here, why do we only care if `_next` is a broadcast stream, shouldn't most of these be checking for `currentStream.isBroadcast`?

test/concat_test.dart
```diff
@@ -8,40 +8,121 @@ import 'package:test/test.dart';
...
+        StreamSubscription subscription;
+
+        setUp(() async {
+          firstCanceled = false;
+          secondCanceled = false;
+          secondListened = false;
+          first = streamTypes[firstType]()
+            ..onCancel = () {
+              firstCanceled = true;
+            };
```
jakemac53, line 37: should we be tracking `onListen` for this as well?

test/concat_test.dart
```diff
@@ -8,40 +8,121 @@ import 'package:test/test.dart';
...
+        });
+
+        test('Does not listen to second stream before first stream finishes',
+            () async {
+          expect(secondListened, false);
+          await first.close();
+          expect(secondListened, true);
+        });
+
+        test('cancels single-subscription first stream on cancel', () async {
```
jakemac53: Should this be inside an `if (firstType == 'single-subscription')` block (the test name indicates that at least).

test/concat_test.dart
```diff
@@ -8,40 +8,121 @@ import 'package:test/test.dart';
...
+          await subscription.cancel();
+          expect(firstCanceled, true);
+        });
+
+        if (firstType == 'single-subscription') {
```
jakemac53: I would expect this to be `secondType == 'single-subscription'` based on the test name?

test/concat_test.dart
```diff
@@ -8,40 +8,121 @@ import 'package:test/test.dart';
...
+          });
+        }
+
+        test('closes stream after both inputs close', () async {
+          await first.close();
+          await second.close();
+          expect(isDone, true);
+        });
+
+        if (firstType == 'broadcast') {
```
jakemac53: it might be worth adding a `group` here with the name `firstType == 'broadcast'` or something

test/concat_test.dart
```diff
@@ -8,40 +8,121 @@ import 'package:test/test.dart';
...
+              .listen(emittedValues.add, onError: errors.add, onDone: () {
+            isDone = true;
+          });
+        });
+
+        test('adds all values from both streams', () async {
+          first..add(1)..add(2);
+          await first.close();
+          await new Future(() {});
+          second..add(3)..add(4);
```
jakemac53, line 59: What if these are added before `first` closes? It seems like in the broadcast stream case you wouldn't catch any of those values?

lib/src/concat.dart
natebosch: The code paths around re-listening should only get hit if `first` is a broadcast stream, otherwise the controller would prevent us from ever getting here.

Our first listener can't hit this since `firstDone` won't ever be true. Subsequent listeners can only hit if `first` was a broadcast stream. We're not still listening to `first` since `firstDone` is true, which means we are (were) listening to `_next` so it's the only stream that `subscription` could be on.

test/concat_test.dart
natebosch: I didn't think it was critical since the test that values come through prove we've listened when we should.

test/concat_test.dart
natebosch: Correct. A lot of behavior in this transformer don't make much sense for broadcast streams, especially so for mixing them, but I thought it'd be best to handle it as well as possible. I can expand the doc comment to call some of this stuff out.

test/concat_test.dart
natebosch: I might need to rename it. If the first stream is broadcast then a single-subscription second stream never gets canceled - only paused.

natebosch (commented):
One thing I'm not sure of - If a single subscription stream is `concat`ed with a broadcast stream, then a cancel will cause it to `pause` - buffer values until it get's listened again.

Maybe instead I should replace the listener with something that ignores values instead?

test/concat_test.dart
jakemac53: Technically if it is a broadcast stream it could start collecting things right away, and just not report them until the first stream is done right? I don't know if that is what users would expect or not.

natebosch:
Come to think of it - what I did with `merge` was to use `asBroadcastStream()` for any single-subscription stuff that get's merged in... maybe that's what I should do here.

test/concat_test.dart
natebosch: The current group names both types of streams so I don't think it would help the output. IMO the group would add a level of indent without much benefit.

lib/src/concat.dart
natebosch: Using `.asBroadcastStream()` allowed me to drop this block

test/concat_test.dart
natebosch: Expanded the doc comment a bit.

natebosch: Somehow I missed the reply earlier.

I don't think it would be best to buffer - it's a potential memory leak. If we have to pick one of 2 bad behaviors I think missing earlier events is the less bad choice.

It's also possible to get that behavior manually by forcing back to single-subscription. Whereas it's not possible to get back to the other behavior if we were to buffer.

```dart
Stream<T> singleSubscriptionEarlyListen<T>(Stream<T> values) {
  var controller = new StreamController<T>();
  values.listen(controller.add, onError: controller.adError, onDone: controller.close);
  return controller.stream;
}
//...
someStream.transform(concat(singleSubscriptionEarlyListen(someBroadcast)));
```

natebosch:
@jakemac53 any other thoughts?

jakemac53 (approved):

PR #33, opened 2017-07-02
Fix asyncWhere for broadcast streams

natebosch:
Previous implementation may have failed to send onDone for some
listeners since each listener would increment `valuesWaiting` and only
the last listener would decrement it back to zero. There were also
potential bugs if the predicate was non-deterministic.


natebosch:
This should be the last thing needed for 0.0.6

grouma (approved):

PR #35, opened 2017-07-12
Add fromBind to replace direct StreamTransformer

### LINKED ISSUE #34: scan and switchMap don't correctly report stream type
`new StreamTransformer(onListen)` constructor doesn't create streams which know their type, so `isBroadcast` always reports `false`. This pattern is used in `scan` and `switchMap` so these produce Streams which don't behave correctly.

### LINKED ISSUE #1756: scan and switchMap don't correctly report stream type
`new StreamTransformer(onListen)` constructor doesn't create streams which know their type, so `isBroadcast` always reports `false`. This pattern is used in `scan` and `switchMap` so these produce Streams which don't behave correctly.

natebosch:
Fixes dart-lang/tools#1756

`new StreamTransformer(onListen)` produces a Stream which can not know
whether it is a Broadcast stream. The `bind` method can more directly
express the intent of these transformers but adding a class is a lot of
boilerplate. Add fromBind to handle these cses.

### DIFF
--- a/lib/src/bind.dart
+++ b/lib/src/bind.dart
@@ -1,0 +1,15 @@
+import 'dart:async';
+
+typedef Stream<T> Bind<S, T>(Stream<S> values);
+
+StreamTransformer<S, T> fromBind<S, T>(Bind<S, T> bindFn) =>
+    new _StreamTransformer(bindFn);
+
+class _StreamTransformer<S, T> implements StreamTransformer<S, T> {
+  final Bind<S, T> _bind;
+
+  _StreamTransformer(this._bind);
+
+  @override
+  Stream<T> bind(Stream<S> values) => _bind(values);
+}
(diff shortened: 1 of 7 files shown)


lib/src/bind.dart
```diff
@@ -0,0 +1,15 @@
+import 'dart:async';
+
+typedef Stream<T> Bind<S, T>(Stream<S> values);
```
grouma: I'd add some documentation.

test/switch_test.dart
```diff
@@ -136,7 +136,14 @@ void main() {
       await new Future(() {});
       outer.add([4, 5, 6]);
       await new Future(() {});
-      expect(values, [1, 2, 3, 4, 5, 6]);
```
grouma: Why was this expect removed?

lib/src/bind.dart
natebosch: Done

test/switch_test.dart
natebosch: Oops. Didn't mean to. Added back.

grouma (approved):

PR #38, opened 2017-08-04
Add takeUntil

### LINKED ISSUE #37: Add takeUntil
```dart
StreamTransformer<T,T> takeUntil<T>(Future end);
```

Firing this future would differ from canceling a subscription in that further transformations which introduce a delay would not be impacted by the moment where the `end` fires, so potentially more events will go through. For example see https://dartpad.dartlang.org/51b9573b2352926cc341f549881ec116

### LINKED ISSUE #1757: Add takeUntil
```dart
StreamTransformer<T,T> takeUntil<T>(Future end);
```

Firing this future would differ from canceling a subscription in that further transformations which introduce a delay would not be impacted by the moment where the `end` fires, so potentially more events will go through. For example see https://dartpad.dartlang.org/51b9573b2352926cc341f549881ec116

natebosch:
Closes dart-lang/tools#1757

### DIFF
--- a/lib/src/take_until.dart
+++ b/lib/src/take_until.dart
@@ -33,7 +33,7 @@ class _TakeUntil<T> implements StreamTransformer<T, T> {
     });
 
     controller.onListen = () {
-      if (subscription != null || isDone) return;
+      if (isDone) return;
       subscription = values.listen(controller.add, onError: controller.addError,
           onDone: () {
         if (isDone) return;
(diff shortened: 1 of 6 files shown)


lib/src/take_until.dart
```diff
@@ -0,0 +1,56 @@
...
+    var isDone = false;
+    _trigger.then((_) {
+      if (isDone) return;
+      isDone = true;
+      subscription?.cancel();
+      controller.close();
+    });
+
+    controller.onListen = () {
+      if (subscription != null || isDone) return;
```
grouma: Shouldn't we throw if the subscription is not null?

natebosch: I'm not sure it's worth checking and throwing - but you're right that it shouldn't happen. This turns out to be a redundant check so I removed it.

grouma (approved):

PR #45, opened 2018-05-24
Explicitly handle null callback for tap

natebosch:
This does not change visible behavior since a null callback would have
previously been an ignored exception.

Rename the argument to `onValue` for consistency and expand the doc
comment.

### DIFF
--- a/lib/src/tap.dart
+++ b/lib/src/tap.dart
@@ -8,16 +8,25 @@ import 'from_handlers.dart';
 /// Taps into a Stream to allow additional handling on a single-subscriber
 /// stream without first wrapping as a broadcast stream.
 ///
-/// The callback will be called with every value from the stream before it is
-/// forwarded to listeners on the stream. Errors from the callbacks are ignored.
+/// The [onValue] callback will be called with every value from the original
+/// stream before it is forwarded to listeners on the resulting stream. May be
+/// null if only [onError] or [onDone] callbacks are needed.
 ///
-/// The tapped stream may not emit any values until the result stream has a
-/// listener, and may be canceled only by the listener.
-StreamTransformer<T, T> tap<T>(void fn(T value),
+/// The [onError] callback will be called with every error from the original
+/// stream before it is forwarded to listeners on the resulting stream.
+///
+/// The [onDone] callback will be called after the original stream closes and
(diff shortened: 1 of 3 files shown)


lib/src/tap.dart
```diff
@@ -8,16 +8,25 @@ import 'from_handlers.dart';
...
+/// stream before it is forwarded to listeners on the resulting stream.
+///
+/// The [onDone] callback will be called after the original stream closes and
+/// before the resulting stream is closed.
+///
+/// Errors from any of the callbacks are ignored.
+///
+/// The callbacks may not be called unil the tapped stream has a listener, and
+/// may not be called after the listener has canceled the subscription.
+StreamTransformer<T, T> tap<T>(void onValue(T value),
```
jakemac53: If we are doing a breaking change anyways why not just change it to an optional arg to indicate that its optional?

natebosch: It's not a breaking change currently. Also by being non-optional it matches a call to stream.listen which also allows explicitly null value callback

natebosch: Ah I didn't realize I had changed the pubspec incorrectly...

This is what I stopped using 0.0.x versioning in other packages :P

jakemac53: I am fine with leaving it as is if the goal is to have a non-breaking change.

natebosch: I double checked and the version bump I have in the pubspec is nonbreaking.

```dart
new Version.parse('0.0.11').nextBreaking; // Results in 0.1.0
```

jakemac53: that doesn't make sense... if I were depending on this package I would be putting `^0.0.11` so this would still cause churn for most users I would think, regardless of what the `nextBreaking` getter says.

natebosch: If your constraint is `^0.0.11` then you can pub solve to `0.0.12`.

As it's published I can put `^0.0.10` in my pubspec and when I `pub get` I see that I'm using `0.0.11`

jakemac53: > As it's published I can put ^0.0.10 in my pubspec and when I pub get I see that I'm using 0.0.11

....... lol well ok then

jakemac53 (approved):

PR #47, opened 2018-06-01
Add `chain` and `forMap` utilities

natebosch:
These are a bit different than others in this package since we would
generally not expect to see them directly inlined to a call to
`transform`, but are instead useful when dealing with variable or
arguments of type `StreamTransformer`, for instance `new
StreamSinkTransformer.fromStreamTransformer`.


natebosch:
I'm very open to other name suggestions, and also open to arguments that these aren't worth publishing.

Example use cases:
chain - https://github.com/dart-lang/test/pull/846/files#diff-1809444246e4db866a7cbd2139e00d63R27
forMap - https://github.com/dart-lang/test/pull/846/files#diff-1809444246e4db866a7cbd2139e00d63R37

grouma (approved):

PR #50, opened 2018-06-13
Add `concurrentAsyncMap`

natebosch:
The `Stream.asyncMap` call is always serial and preserves ordering.

### DIFF
--- a/test/concurrent_async_map_test.dart
+++ b/test/concurrent_async_map_test.dart
@@ -1,0 +1,160 @@
+// Copyright (c) 2018, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+
+import 'package:test/test.dart';
+
+import 'package:stream_transform/stream_transform.dart';
+
+void main() {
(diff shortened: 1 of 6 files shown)


test/concurrent_async_map_test.dart
```diff
@@ -0,0 +1,160 @@
...
+        await new Future(() {});
+        expect(emittedValues, [1]);
+      });
+
+      test('allows calls to convert before the last one finished', () async {
+        streamController.add(1);
+        streamController.add(2);
+        streamController.add(3);
+        await new Future(() {});
+        expect(values, [1, 2, 3]);
```
grouma, line 70: How is this test ensuring the convert can be called before the last one? Is it implicit that they are completed before a single pass through the event loop?

grouma (approved):

test/concurrent_async_map_test.dart
natebosch: items aren't added to `values` until the convert callback is called, and the convert callback doesn't finish until you call `finishWork[x].complete()` so this is indicating that 3 values have been passed to `convert` before any value has finished.

grouma: Ahh I see now.

PR #55, opened 2018-08-06
Run dartfmt --fix to drop optional `new`

natebosch:


jakemac53 (approved):
Looks like you need to update the package:test dependency, but lgtm

PR #60, opened 2019-02-11
Add whereType transformer

natebosch:
This needs to be implemented as it's own class instead of using
one of the helpers like `fromHandlers` because it has unique generic
type requirements. We want to avoid requring the call site to re-specify
the generic type of the input stream. The return type needs to be a
`StreamTransformer<Null, R>` to allow to satisfy static checking.
However the argument to `bind` at runtime will never be a `Stream<Null>`
so we need to define it explicitly to widen the allowed argument.

### DIFF
--- a/lib/src/where_type.dart
+++ b/lib/src/where_type.dart
@@ -9,6 +9,14 @@ import 'dart:async';
 /// If the source stream is a broadcast stream the result will be as well.
 ///
 /// Errors from the source stream are forwarded directly to the result stream.
+///
+/// The static type of the returned transformer takes `Null` so that it can
+/// satisfy the subtype requirements for `stream.transform()` argument on any
+/// source Stream. The argument to `bind` has been broaded to take
+/// `Stream<Object>` since it never be passed a `Stream<Null>` at runtime. This
+/// is safe to use on any source stream and there is no static or runtime
+/// checking that [R] is sensible - that is that is a subtype of the stream's
+/// type such that some values of that type may be possible.
 StreamTransformer<Null, R> whereType<R>() => _WhereType<R>();
 
 class _WhereType<R> extends StreamTransformerBase<Null, R> {
--- a/lib/src/where_type.dart
+++ b/lib/src/where_type.dart
@@ -21,14 +21,13 @@ class _WhereType<R> extends StreamTransformerBase<Null, R> {
     StreamSubscription<Object> subscription;
     controller.onListen = () {
       if (subscription != null) return;
-      var valuesDone = false;
       subscription = values.listen(
           (value) {
             if (value is R) controller.add(value);
           },
           onError: controller.addError,
           onDone: () {
-            valuesDone = true;
+            subscription = null;
             controller.close();
           });
       if (!values.isBroadcast) {
@@ -36,10 +35,8 @@ class _WhereType<R> extends StreamTransformerBase<Null, R> {
         controller.onResume = subscription.resume;
       }
       controller.onCancel = () {
-        var toCancel = subscription;
+        subscription?.cancel();
         subscription = null;
-        if (!valuesDone) return toCancel.cancel();
-        return null;
       };
     };
     return controller.stream;
--- a/lib/src/where_type.dart
+++ b/lib/src/where_type.dart
@@ -16,7 +16,7 @@ import 'dart:async';
 /// `Stream<Object>` since it will never be passed a `Stream<Null>` at runtime.
 /// This is safe to use on any source stream.
 ///
-/// [R] should b a subtype of the stream's generic type otherwise nothing of
+/// [R] should be a subtype of the stream's generic type, otherwise nothing of
 /// type [R] could possibly be emitted, however there is no static or runtime
 /// checking that this is the case.
 StreamTransformer<Null, R> whereType<R>() => _WhereType<R>();
--- a/lib/src/where_type.dart
+++ b/lib/src/where_type.dart
@@ -11,25 +11,27 @@ import 'dart:async';
 /// Errors from the source stream are forwarded directly to the result stream.
 ///
 /// The static type of the returned transformer takes `Null` so that it can
-/// satisfy the subtype requirements for `stream.transform()` argument on any
-/// source Stream. The argument to `bind` has been broaded to take
-/// `Stream<Object>` since it never be passed a `Stream<Null>` at runtime. This
-/// is safe to use on any source stream and there is no static or runtime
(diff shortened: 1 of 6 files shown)


lib/src/where_type.dart
```diff
@@ -0,0 +1,47 @@
...
+import 'dart:async';
+
+/// Emits only the events which have type [R].
+///
+/// If the source stream is a broadcast stream the result will be as well.
+///
+/// Errors from the source stream are forwarded directly to the result stream.
+StreamTransformer<Null, R> whereType<R>() => _WhereType<R>();
+
+class _WhereType<R> extends StreamTransformerBase<Null, R> {
```
jakemac53, line 24: The null is a code smell here, what does it represent? can it also be generic, or should it maybe be Object?

lib/src/where_type.dart
```diff
@@ -0,0 +1,47 @@
...
+  @override
+  Stream<R> bind(Stream<Object> values) {
```
jakemac53: nit: maybe `inputs` instead of values? values is fairly ambiguous

lib/src/where_type.dart
```diff
@@ -0,0 +1,47 @@
...
+            controller.close();
+          });
+      if (!values.isBroadcast) {
+        controller.onPause = subscription.pause;
+        controller.onResume = subscription.resume;
+      }
+      controller.onCancel = () {
+        var toCancel = subscription;
+        subscription = null;
+        if (!valuesDone) return toCancel.cancel();
```
jakemac53: It seems like it would be cleaner to null out the `subscription` field when the values are done, and get rid of this `valuesDone` field entirely.

lib/src/where_type.dart
jakemac53: Oh I see you commented about this, I don't fully understand the issue here but I can check out the repo and try it out which would probably make it more obvious.

jakemac53:
As a more general comment, stream does already have a `where` method... a stream transformer seems quite heavy for this versus just doing `stream.where((thing) => thing is SomeType)`?

lib/src/where_type.dart
natebosch: See some discussion here:
https://github.com/dart-lang/language/issues/180#issuecomment-461998427

concrete scenario:

```dart
  var numbers = Stream<num>.fromIterable([1, 1.5, 2]);
  var ints = numbers.transform(whereType<int>());
```

In this case the `transform` excepts an argument assignable to `StreamTransformer<num, S>`. Since the `whereType` call doesn't have a way to statically know the `num` there, it either needs us to be explicit about it (`numbers.transform(whereType<num, int>()`), or to use some default that can fill in for any `T` that may have been on the stream. `Null` is the only possibility there for now, I'm not sure if we've figured out what the user referenceable  bottom type will be after NNBD...

natebosch:
> a stream transformer seems quite heavy for this versus just doing `stream.where((thing) => thing is SomeType)`?

That doesn't end up with the correct static or runtime type. You either need to chain a `.cast()` call, or use a hack like `stream.expand<SomeType>((thing) => thing is SomeType ? [thing] : [])`. Both of these incur some amount of additional runtime overhead, either to construct a list or to do the cast that will be guaranteed to succeed.

lib/src/where_type.dart
jakemac53: I would at least leave a lengthy doc comment here, its going to be pretty bizarre for anybody reading this in the future so some context would certainly help :).

lib/src/where_type.dart
natebosch: This matches the naming pattern used elsewhere in the package.
https://github.com/dart-lang/stream_transform/blob/5c5049333dcef448acebb09c6d53b6c99c1b99e8/lib/src/from_handlers.dart#L50

jakemac53: Do you think its actually a good name though?

lib/src/where_type.dart
natebosch: I think I was worried about the `onListen` getting called again in a broadcast stream after the values where already done and the `subscription != null` condition not blocking us from doing work we shouldn't - but it looks like maybe `StreamController` handles this for us and won't call `onListen` again anyway. Done

lib/src/where_type.dart
natebosch: I think it's about as good as anything else. A `value` or `event` is what is emitted by a stream when we don't know anything more specific about it, and I generally like to name a stream as the plural of whatever flows through it. I don't like the name `inputs` because I wouldn't call the thing emitted by the stream an `input` - I wouldn't name the argument to the listen callback that.

lib/src/where_type.dart
natebosch: Added a note to the doc comment, how does that look to you? I think for further details the `git blame` should point here which will have extra detail.

jakemac53: I think what you added is fine

lib/src/where_type.dart
jakemac53: > I don't like the name `inputs` because I wouldn't call the thing emitted by the stream an `input`

In the case of a stream transformer I would disagree - it is transforming one stream into another stream, so it makes sense to me to consider one the "inputs" stream and one the "outputs" stream.

If I see the word "values" in the middle of the method its not clear what that represents. Is that the outputs or the inputs? But if I see "inputs" I can immediately assume its the input stream.

In any case I don't care enough to withhold an lgtm, but I do think "inputs" is better :).

> I wouldn't name the argument to the listen callback that.

Right, because listen doesn't transform anything, it only listens for values and there is only one stream in scope. With a transformer there are two streams in the scope of the `bind` method.

lib/src/where_type.dart
```diff
@@ -21,25 +21,22 @@ class _WhereType<R> extends StreamTransformerBase<Null, R> {
...
+            subscription = null;
             controller.close();
           });
       if (!values.isBroadcast) {
         controller.onPause = subscription.pause;
         controller.onResume = subscription.resume;
       }
       controller.onCancel = () {
-        var toCancel = subscription;
+        subscription?.cancel();
```
jakemac53, line 48: this used to return this value, not sure if it matters or not

jakemac53 (approved):

lib/src/where_type.dart
natebosch: Hmm, I'm not sure either. There was some weirdness with an analyzer hint around this at one point that could have been why I had it in the `fromHandlers` implementation, and I copied most of that here.

PR #64, opened 2019-03-08
Allow a `FutureOr` return in `scan`

### LINKED ISSUE #62: Add asyncScan
Hi,

I am wondering whether it would make sense to add an `asyncScan` method to this library.

The idea would be to build a kind of reducer that provides the next value asynchronously.

My use case would be the data source of an infinitely scrolling list, where the UI passes a watermark and the stream would fetch the next items off the network and return a new `List<ViewModel>`.

Would this be a useful addition, or is there another construct that would be better suited to this case?

Thanks!

### LINKED ISSUE #1761: Add asyncScan
Hi,

I am wondering whether it would make sense to add an `asyncScan` method to this library.

The idea would be to build a kind of reducer that provides the next value asynchronously.

My use case would be the data source of an infinitely scrolling list, where the UI passes a watermark and the stream would fetch the next items off the network and return a new `List<ViewModel>`.

Would this be a useful addition, or is there another construct that would be better suited to this case?

Thanks!

natebosch:
Closes dart-lang/tools#1761

Adds async support in `scan` instead of a new `asyncScan` since the
latter would be able to serve both use cases so it would not be clear
why both exist.

Technically this is breaking but it's such an extreme edge case that it
is very unlikely to come up in practice - and the breakage is detected
statically. Due to the way `FuturOr<Future>` interacts with inference
there are some cases where the generic types may need to be explicitly
written. This is an unlikely scenario where the transformer emitted a
stream of futures.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## 0.0.16
+
+- Allow a `combine` callback to return a `FutureOr<T>` in `scan`. There are no
+  behavior changes for synchronous callbacks. **Potential breaking change** In
+  the unlikely situation where `scan` was used to produce a `Stream<Future>`
+  inference may now fail and require explicit generic type arguments.
+
 ## 0.0.15
 
 - Add `whereType`.
--- a/lib/src/scan.dart
+++ b/lib/src/scan.dart
@@ -6,9 +6,21 @@ import 'dart:async';
 
 /// Scan is like fold, but instead of producing a single value it yields
 /// each intermediate accumulation.
+///
+/// If [combine] returns a Future it will not be called again for subsequent
+/// events from the source until it completes, therefor the combine callback is
+/// always called for elements in order, and the result stream always maintains
+/// the same order as the original.
 StreamTransformer<S, T> scan<S, T>(
-        T initialValue, T combine(T previousValue, S element)) =>
-    StreamTransformer.fromBind((stream) {
+        T initialValue, FutureOr<T> combine(T previousValue, S element)) =>
+    StreamTransformer.fromBind((source) {
       var accumulated = initialValue;
-      return stream.map((value) => accumulated = combine(accumulated, value));
+      return source.asyncMap((value) {
+        var result = combine(accumulated, value);
+        if (result is Future<T>) {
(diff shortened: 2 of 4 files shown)


jakemac53 (approved):

tp:
Thanks for implementation this so quickly :)

I added the latest version to my project to try it out:

I have code similar to this:
```dart
final firstProductIndex = StreamController<int>();

firstProductIndex.map((firstProductIndex) => 1 /* page to load… */).transform(
        scan<int, int>(0, (previous, element) => max(previous, element)));
```

As you can see, for the `scan` I had to manually add type annotations, otherwise the initial value would be inferred as dynamic.

Is that a regression / could that be improved to not require an annotation?

Thanks!

natebosch:
> I added the latest version to my project to try it out:

Using a git dependency? It's not published with the `FutureOr` change yet.

>I had to manually add type annotations, otherwise the initial value would be inferred as dynamic.
>
> Is that a regression / could that be improved to not require an annotation?

I don't think it's a regression. I'm guessing here your `max` is the one from `dart:math`? If that's the case the `FutureOr` return isn't making a difference here. The problem is that `max` is also generic.
https://api.dartlang.org/stable/2.2.0/dart-math/max.html
That's where you see the analysis error right?

Inference is not helping you because you don't have enough to hint at the return type. `Object` could get filled in as the return type for the result stream, but that won't work for the `T` in `max` - though it does work for both `0` and the argument types for `previous` and `element`. Inference can be hard to follow because it goes both "down" into the method, as well as "up" from the return type on the method.

You have to fill it in somewhere. I'd recommend typing the arguments `previous` and `element`, but you can also fill it in with `scan<int, int>` as you've discovered, or to `max<int>` if you prefer.

```dart
final firstProductIndex = StreamController<int>();

firstProductIndex.map((firstProductIndex) => 1 /* page to load… */).transform(
        scan(0, (int previous, int element) => max(previous, element)));
```

tp:
> Using a git dependency? It's not published with the `FutureOr` change yet.

Yes :)


>  I'd recommend typing the arguments `previous` and `element`

Sweet, that is so much nicer to follow.

Thanks!

PR #65, opened 2019-03-08
Add `combineLatest`

### LINKED ISSUE #63: Implement a `combineLatest` transform
I'd expect usage to look like this:

```dart
firstStream.transform(combineLatest(secondStream, (a, b) => a + b));
```

I'd expect us to _not_ pass `null` to the combine in case one of the streams hasn't emitted yet - to get that behavior you'd could use a `startWith` to force an "immediate" start for one or both streams.

### LINKED ISSUE #1762: Implement a `combineLatest` transform
I'd expect usage to look like this:

```dart
firstStream.transform(combineLatest(secondStream, (a, b) => a + b));
```

I'd expect us to _not_ pass `null` to the combine in case one of the streams hasn't emitted yet - to get that behavior you'd could use a `startWith` to force an "immediate" start for one or both streams.

natebosch:
Closes dart-lang/tools#1762

Uses an manual `StreamTransformer` implementation based closely on the
`merge` implementation. This can't be implemented on top of
`fromHandlers` or normal stream methods because the events emitted
aren't triggered by a single source.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,11 @@ the most recent value.
 Collects values from a source stream until a `trigger` stream fires and the
 collected values are emitted.
 
+# combineLatest
+
+Combine the most recent event from two streams through a callback and emit thre
+result.
+
 # debounce, debounceBuffer
 
 Prevents a source stream from emitting too frequently by dropping or collecting
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ collected values are emitted.
 
 # combineLatest
 
-Combine the most recent event from two streams through a callback and emit thre
+Combine the most recent event from two streams through a callback and emit the
 result.
 
 # debounce, debounceBuffer
--- a/lib/src/combine_latest.dart
+++ b/lib/src/combine_latest.dart
@@ -1,0 +1,119 @@
+// Copyright (c) 2019, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+
+/// Combine the latest value from the source stream with the latest value from
+/// [combineWith] using [combine].
+///
+/// No event will be emitted from the result stream until both the source stream
+/// and [combineWith] have each emitted at least one event. The result stream
+/// will not close until both the source stream and [combineWith] have closed.
+/// Errors throw by [combine], along with any errors on the source stream or
+/// [combineWith], are forwarded to the result stream.
+///
+/// If the source stream is a broadcast stream, the result stream will be as
+/// well, regardless of [combineWith]'s type. If a single subscription stream is
+/// merged into a broadcast stream it may never be canceled.
+StreamTransformer<S, R> combineLatest<S, T, R>(
+        Stream<T> combineWith, FutureOr<R> Function(S, T) combine) =>
+    _CombineLatest(combineWith, combine);
+
+class _CombineLatest<S, T, R> extends StreamTransformerBase<S, R> {
+  final Stream<T> _other;
+  final FutureOr<R> Function(S, T) _combine;
+
+  _CombineLatest(this._other, this._combine);
+
+  @override
+  Stream<R> bind(Stream<S> source) {
+    final controller = source.isBroadcast
+        ? StreamController<R>.broadcast(sync: true)
+        : StreamController<R>(sync: true);
+
+    final other = (source.isBroadcast && !_other.isBroadcast)
+        ? _other.asBroadcastStream()
+        : _other;
+
+    StreamSubscription sourceSubscription;
+    StreamSubscription otherSubscription;
+
+    var sourceDone = false;
+    var otherDone = false;
+
+    S latestSource;
+    T latestOther;
+
+    var sourceStarted = false;
+    var otherStarted = false;
+
+    void emitCombined() {
+      if (!sourceStarted || !otherStarted) return;
+      FutureOr<R> result;
+      try {
+        result = _combine(latestSource, latestOther);
+      } catch (e, s) {
+        controller.addError(e, s);
+        return;
+      }
+      if (result is Future<R>) {
+        result
+            .then(controller.add, onError: controller.addError)
+            .whenComplete(() {
+          sourceSubscription.resume();
+          otherSubscription.resume();
+        });
+      } else {
+        controller.add(result as R);
+      }
+    }
+
+    controller.onListen = () {
+      if (sourceSubscription != null) return;
+      sourceSubscription = source.listen(
+          (s) {
+            sourceStarted = true;
+            latestSource = s;
+            emitCombined();
+          },
+          onError: controller.addError,
+          onDone: () {
+            sourceDone = true;
+            if (otherDone) controller.close();
+          });
+      otherSubscription = other.listen(
+          (o) {
+            otherStarted = true;
+            latestOther = o;
+            emitCombined();
+          },
+          onError: controller.addError,
+          onDone: () {
+            otherDone = true;
+            if (sourceDone) controller.close();
+          });
+      if (!source.isBroadcast) {
+        controller
+          ..onPause = () {
+            sourceSubscription.pause();
+            otherSubscription.pause();
+          }
+          ..onResume = () {
+            sourceSubscription.resume();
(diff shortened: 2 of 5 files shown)


lib/src/combine_latest.dart
```diff
@@ -0,0 +1,119 @@
...
+          };
+      }
+      controller.onCancel = () async {
+        var cancelSource = sourceSubscription.cancel();
+        var cancelOther = otherSubscription.cancel();
+        sourceSubscription = null;
+        otherSubscription = null;
+
+        await cancelSource;
+        await cancelOther;
```
lrhn: Here you await two futures in order. If the second future completes with an error before the first one completes at all, that error will be uncaught.

Consider doing `await Future.wait([cancelSource, cancelOther]);`.

natebosch: Good catch. Thanks

README.md
```diff
@@ -19,6 +19,11 @@ the most recent value.
 Collects values from a source stream until a `trigger` stream fires and the
 collected values are emitted.
 
+# combineLatest
+
+Combine the most recent event from two streams through a callback and emit thre
```
TastyPi: Typo: "thre"

lib/src/combine_latest.dart
```diff
@@ -0,0 +1,129 @@
...
+
+import 'dart:async';
+
+/// Combine the latest value from the source stream with the latest value from
+/// [combineWith] using [combine].
+///
+/// No event will be emitted from the result stream until both the source stream
+/// and [combineWith] have each emitted at least one event. The result stream
+/// will not close until both the source stream and [combineWith] have closed.
+/// Errors throw by [combine], along with any errors on the source stream or
```
TastyPi: Typo: "Errors throw" should be "Errors thrown"

lib/src/combine_latest.dart
```diff
@@ -0,0 +1,129 @@
...
+/// [combineWith], are forwarded to the result stream.
+///
+/// If the source stream is a broadcast stream, the result stream will be as
+/// well, regardless of [combineWith]'s type. If a single subscription stream is
+/// merged into a broadcast stream it may never be canceled.
```
TastyPi: nit: "merge" is a different operator, maybe "a single subscription stream is combined with"

lib/src/combine_latest.dart
```diff
@@ -0,0 +1,129 @@
...
+      } catch (e, s) {
+        controller.addError(e, s);
+        return;
+      }
+      if (result is Future<R>) {
+        result
+            .then(controller.add, onError: controller.addError)
+            .whenComplete(() {
+          sourceSubscription.resume();
+          otherSubscription.resume();
```
TastyPi, line 82: Should these subscriptions have been paused first?

test/combine_latest_test.dart
```diff
@@ -0,0 +1,137 @@
...
+      var combineWith = StreamController<int>();
+      int sum(int a, int b) => a + b;
+
+      var results = <int>[];
+      unawaited(source.stream
+          .transform(combineLatest(combineWith.stream, sum))
+          .forEach(results.add));
+
+      source.add(1);
+      await Future(() {});
```
TastyPi, line 25: Why not pumpEventQueue? https://docs.flutter.io/flutter/package-test_api_test_api/pumpEventQueue.html

test/combine_latest_test.dart
```diff
@@ -0,0 +1,137 @@
...
+      await source.close();
+      await Future(() {});
+      expect(done, false);
+
+      await combineWith.close();
+      await Future(() {});
+      expect(done, true);
+    });
+
+    test('ends if a Stream closes without ever emitting a value', () async {
```
TastyPi: Should test the source stream being empty too.

lib/src/combine_latest.dart
```diff
@@ -0,0 +1,129 @@
...
+  _CombineLatest(this._other, this._combine);
+
+  @override
+  Stream<R> bind(Stream<S> source) {
+    final controller = source.isBroadcast
+        ? StreamController<R>.broadcast(sync: true)
+        : StreamController<R>(sync: true);
+
+    final other = (source.isBroadcast && !_other.isBroadcast)
+        ? _other.asBroadcastStream()
```
jakemac53, line 51: What is the purpose of converting this to a broadcast stream internally?

lib/src/combine_latest.dart
```diff
@@ -0,0 +1,129 @@
...
+          sourceSubscription.resume();
+          otherSubscription.resume();
+        });
+      } else {
+        controller.add(result as R);
+      }
+    }
+
+    controller.onListen = () {
+      if (sourceSubscription != null) return;
```
jakemac53: Afaik this check isn't strictly necessary, you could convert it to an assert.

test/combine_latest_test.dart
```diff
@@ -0,0 +1,137 @@
...
+
+import 'dart:async';
+
+import 'package:pedantic/pedantic.dart';
+import 'package:test/test.dart';
+
+import 'package:stream_transform/stream_transform.dart';
+
+void main() {
+  group('combineLatest', () {
```
jakemac53, line 13: nit: It would probably be good to have at least one test which combines two strings with different types.

lib/src/combine_latest.dart
```diff
@@ -0,0 +1,129 @@
+// Copyright (c) 2019, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+
+/// Combine the latest value from the source stream with the latest value from
+/// [combineWith] using [combine].
```
jakemac53: nit: add a clarifying note here that this emits a value any time either stream emits a value as opposed to waiting for a new value from each stream before emitting a new value.

I think it might be worth even listing a quick example similar to how the first test looks.

lib/src/combine_latest.dart
```diff
@@ -0,0 +1,129 @@
...
+///
+/// No event will be emitted from the result stream until both the source stream
+/// and [combineWith] have each emitted at least one event. The result stream
```
jakemac53: I would separate these all out into their own paragraphs (or bullets?) - they are all descriptions of distinct behaviors.

README.md
natebosch: Fixed

lib/src/combine_latest.dart
natebosch: Added another sentence and a timeline diagram - does that make it clear?

lib/src/combine_latest.dart
natebosch: Done

lib/src/combine_latest.dart
natebosch: Fixed

lib/src/combine_latest.dart
natebosch: Done

lib/src/combine_latest.dart
natebosch: It's the only way we can cancel and relisten. The other option would be to replace the listener with something that throws away values while it's canceled... but I think that's about what `asBroadcastStream` does for us.

lib/src/combine_latest.dart
natebosch: Yes they should have been. Fixed.

lib/src/combine_latest.dart
natebosch: Done.

test/combine_latest_test.dart
natebosch: Done

test/combine_latest_test.dart
natebosch: `pumpEventQueue` is good for the cases where the work that's happening may create additional trips around the event loop while it's running. In this case we wouldn't want that to happen (though, admittedly, we aren't really being precise about guaranteeing it)

I'll leave this as is for now, if we decide to switch over to `pumpEventQueue` we'll probably do all the tests at once.

test/combine_latest_test.dart
natebosch: Done

jakemac53 (approved):

PR #67, opened 2019-03-11
Add `combineLatestAll`

brettapeters:
Adds a `combineLatestAll` transform, which is similar to `combineLatest`, but combines an iterable of Streams together and emits a list of latest values.


natebosch (approved):

natebosch:
Looks great. Thanks!

PR #69, opened 2019-03-12
More lints

natebosch:


analysis_options.yaml
```diff
@@ -3,6 +3,7 @@ analyzer:
   strong-mode:
     implicit-casts: false
   errors:
+    todo: ignore
```
kevmoo, line 6: does `todo` come up?

kevmoo (approved):
I'm on the fence about cascades, but if you like it...

analysis_options.yaml
natebosch: it hasn't yet in this package but I'm in the habit of adding this line anytime I'm doing a bigger cleanup.

natebosch:
> I'm on the fence about cascades, but if you like it...

Yeah, some of the cases for this lint I like and some I could go either way. We're trying it in the build repo and I copied the list of lints over from there. I figured I may as well keep it until we find an egregious case or learn that its too annoying day to day.

PR #75, opened 2019-03-20
Make tap's onError arguments have specific types

TastyPi:
Currently `onError` has type `void Function(dynamic, dynamic)`.

This causes a problem with something like:

```
void myOnError(Object error, StackTrace stackTrace) {}
stream.transform(tap(null, onError: myOnError)).listen(null);
```

As that causes an error:

```
type '(Object, StackTrace) => void' is not a subtype of type '(dynamic, dynamic) => void'
```

Note that `(dynamic, dynamic) => void` is a subtype of `(Object, StackTrace) => void`, so any function that can be passed to the old version can also be passed to the new one.


natebosch (approved):
Thanks!

PR #78, opened 2019-04-26
Enable and fix strict-raw-types checks

natebosch:
Most changes are in tests where we implicitly were using `<dynamic>` but
only ever passing a single type. Add specific types and clean up the few
cases where a different type from the norm was used. Extract utilities
for constructing both types of stream to do the tests in a loop, since
it's not using function literals anymore it can be a generic function.

In `lib/` add a few `<void>` type arguments for things like "trigger"
streams since we were never reading the value.

### DIFF
--- a/lib/src/async_map_buffer.dart
+++ b/lib/src/async_map_buffer.dart
@@ -31,19 +31,21 @@ StreamTransformer<S, T> asyncMapBuffer<S, T>(
   var workFinished = StreamController()
     // Let the first event through.
     ..add(null);
-  return chainTransformers(
-      buffer(workFinished.stream), _asyncMapThen(convert, workFinished.add));
+  return chainTransformers(buffer(workFinished.stream),
+      _asyncMapThen(convert, () => workFinished.add(null)));
 }
 
 /// Like [Stream.asyncMap] but the [convert] is only called once per event,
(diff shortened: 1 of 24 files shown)


natebosch:
This case feels like a pretty clear improvement to me. The tests were annoying to migrate but I think they are better for it. The change in `lib/` are pretty clearly for the better.

lib/src/async_map_buffer.dart
```diff
@@ -31,19 +31,21 @@ StreamTransformer<S, T> asyncMapBuffer<S, T>(
   var workFinished = StreamController()
     // Let the first event through.
     ..add(null);
-  return chainTransformers(
-      buffer(workFinished.stream), _asyncMapThen(convert, workFinished.add));
+  return chainTransformers(buffer(workFinished.stream),
+      _asyncMapThen(convert, () => workFinished.add(null)));
```
jakemac53: Fwiw is going to be slower than the tearoff, so there is some negative value in this case.

jakemac53 (approved):

lib/src/async_map_buffer.dart
natebosch: Good call. `void Function(void)` argument feels weird, but turns out to be fine for our use case if I go back to `.then` instead of `.whenComplete`.

PR #79, opened 2019-05-03
Add asyncMapSample transform

natebosch:
Like `asyncMapBuffer` but drops instead of collecting previous events.

- Refactor `_Buffer` class into `Aggregate` which takes a function to
  aggregate values instead of only collecting them to a List. This will
  not be published.
- Add `asyncMapSample` which is effectively an identical implementation
  to `asyncMapBuffer` except it chooses a different aggregation
  function.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@ Utility methods to create `StreamTransfomer` instances to manipulate Streams.
 Like `asyncMap` but events are buffered in a List until previous events have
 been processed rather than being called for each element individually.
 
-# asyncMapBuffer
+# asyncMapSample
 
 Like `asyncMap` but events are discarded, keeping only the latest, until
 previous events have been processed rather than being called for every element.
--- a/README.md
+++ b/README.md
@@ -5,6 +5,11 @@ Utility methods to create `StreamTransfomer` instances to manipulate Streams.
 Like `asyncMap` but events are buffered in a List until previous events have
 been processed rather than being called for each element individually.
 
+# asyncMapBuffer
+
+Like `asyncMap` but events are discarded, keeping only the latest, until
+previous events have been processed rather than being called for every element.
+
 # asyncWhere
 
 Like `where` but allows an asynchronous predicate.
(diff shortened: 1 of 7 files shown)


natebosch:
I'm unsure about the naming. `sample` has a meaning in Rx that doesn't quite match this - it's a time based concept, not a "trigger" based concept. I feel like there _should_ be an operator that has the semantics of "emit the latest event whenever a trigger fires", but I can't find it...

kevmoo:
Here's your target:
https://github.com/kevmoo/qr.dart/blob/master/example/src/throttled_stream.dart

*From: *Nate Bosch <notifications@github.com>
*Date: *Fri, May 3, 2019 at 9:52 AM
*To: *dart-lang/stream_transform
*Cc: *Kevin Moore, Review requested

I'm unsure about the naming. sample has a meaning in Rx that doesn't quite

README.md
```diff
@@ -5,6 +5,11 @@ Utility methods to create `StreamTransfomer` instances to manipulate Streams.
 Like `asyncMap` but events are buffered in a List until previous events have
 been processed rather than being called for each element individually.
 
+# asyncMapBuffer
```
kevmoo: this was already here – just documenting it?

natebosch: Copy/paste/fail to rename. 😳

natebosch:
Here's about the closest I can get to that interface using this:

```dart
import 'dart:async';

import 'package:stream_transform/stream_transform.dart';

class ThrottledStream<S, T> {
  final StreamController<S> _events;
  S _source; // Do we need this?
  
  final Stream<T> outputStream;

  factory ThrottledStream(Future<T> Function(S source) convert) {
    final events = StreamController<S>.broadcast();
    final outputStream = events.transform(asyncMapSample(convert));
    return ThrottledStream._(events, outputStream);
    }
  
  ThrottledStream._(this._events, this.outputStream);

  Stream<T> get outputStream => _controller.stream;

  S get source => _source;

  set source(S value) {
    _source = value;
    _events.add(value);
  }

  // Can't really support this...
  //T get outputValue => _outputValue;

  Future close() => _events.close();
}
```

kevmoo (approved):
LGTM! Tried it out. Very nice!

PR #81, opened 2019-07-19
Inline some typedefs

natebosch:
In each place where these were used they matched the argument/field name
anyway so it was repetitive and the name was not adding useful
information. They were all a bit shorter than the function type, but not
so much so that it's worth the extra level of indirection.


kevmoo (approved):
sure...

PR #82, opened 2019-09-04
Feature: Add distinctWhen

felangel:
`distinctWhen` will allow developers to apply the `distinct` operator on a `Stream` conditionally.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,10 @@ list of the values.
 Prevents a source stream from emitting too frequently by dropping or collecting
 values that occur within a given duration.
 
+# distinctWhen
+
+Allows developers to apply the distinct operator on a Stream conditionally.
+
 # concurrentAsyncMap
 
 Like `asyncMap` but the convert callback can be called with subsequent values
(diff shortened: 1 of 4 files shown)


natebosch:
Can you describe your use case?

felangel:
We want to be able to filter a stream of SearchEvents and ignore duplicate searches but allow duplicate RetryEvents.

natebosch:
I wonder if it would be straightforward enough to use the built in `Stream.distict` and implement this with a custom equals check?

```dart
var distinctSearchOrRetry = events.distinct((previous, next) {
  // retry events are always considered distinct
  if (next is RetryEvent) return false;
  return previous == next;
});
```

compared with:

```dart
var distinctSearchOrRetry = events.transform(distinctWhen((e) => e is RetryEvent));
```

It's a little more verbose, but it might sidestep some of the questions around exactly what the meaning of `distinctWhen` is.

felangel:
Sounds good, thanks for the feedback 👍

PR #83, opened 2019-10-02
Rewrite transformers as extension methods

natebosch:
The usage is much nicer as extension methods.

In most cases a small bit of code is copied for the replacement, though
in some cases nearly the entire implementation is copied. This was done
over extracting shared pieces so that the deprecated methods can be
removed outright in the next major version change.

Update and clean up some doc comments in the extension method version of
these functions.

Migrate tests to use the extension method versions. The old copies are
now untested but they won't be modified before they are removed.

Since there aren't useful needs within this package for
`chainTransformers` or `map` they are deprecated without a replacement.


natebosch:
We won't be able to move forward with this until we have at least a dev SDK with extensions enabled. We likely won't publish until it's a stable SDK.

I wanted to get the PR out early for feedback.

natebosch:
One open question I have is how to think about which ones go into the same extension and how to name the extensions. It'll be breaking to change this later because of `show` and `hide` on imports.

I don't _think_ it makes sense to try to merge these all into one extension, so we'll want to consider carefully how to group them.

CHANGELOG.md
```diff
@@ -1,3 +1,22 @@
...
+-   The `map` and `chainTransformers` utilities are no longer useful with the
+    new patterns so they are deprecated without a replacement. If you still have
+    a need for them they can be replicated with `StreamTransformer.fromBind`:
+
+    ```
+    // Replace `map(convert)`
+    StreamTransformer.fromBind((s) => s.map(convert));
+
+    // Replace `chainTransformers(first, second)`
+    StreamTransformer.fromBInd((s) => s.transform(first).transform(second));
```
jakemac53: nit: s/BInd/Bind

lib/src/audit.dart
```diff
@@ -5,6 +5,45 @@ import 'dart:async';
 
 import 'from_handlers.dart';
 
+extension Audit<T> on Stream<T> {
+  /// Returns a Stream which only emits once per [duration], at the end of the
+  /// period.
+  ///
+  /// Always introduces a delay of at most [duration].
```
jakemac53: It's really `at least` not `at most`? Other async/sync work can starve this for an arbitrary amount of time.

jakemac53 (commented):
I didn't look at the impls - I assume they are the same.

Usage sites look great, clear improvement there that justifies the change imo.

As far as one extension versus multiple, I can't think of a reason to really prefer one over the other, except that organizationally within this package it seems nicer to have more targeted, smaller libraries. So I am perfectly fine with how it is currently set up.

lib/src/audit.dart
natebosch: I think it's tricky to describe. "Always introduces a delay of at most [duration] plus arbitrary delay for unrelated async work." might be the most accurate.

Some events won't be emitted. For the events that _are_ emitted, if it was the event that triggered the timer it will be delayed by `duration + arbitrary_async`. If it was an event that came in within that period it's delayed by `(duration - delta_since_event_triggering_timer) + arbitrary_async`...

natebosch: I updated the comment with a longer explanation and a timeline example. Does it make more sense this way?

PR #90, opened 2019-11-27
Add concurrentAsyncExpand

### LINKED ISSUE #77: Add concurrentAsyncExpand?
We have `Stream.asyncMap`, `concurrentAsyncMap`, `Stream.asyncExpand`... but no `concurrentAsyncExpand`.

```dart
StreamTransformer<S, T> concurrentAsyncExpand<S, T>(Stream<T> convert(S event))
```

Example:

```dart
Iterable<Future<int>> futures = [3, 2, 1].map((i) => Future.delayed(Duration(seconds: i), () => i));
Stream.fromIterable(futures)
  .transform(concurrentAsyncExpand((future) async* {
    yield 'awaiting future';
    var result = await future;
    yield 'future completed: $result';
  }))
  .forEach(print);
```

Output:
```
awaiting future
awaiting future
awaiting future
future completed: 1
future completed: 2
future completed: 3
```

### LINKED ISSUE #1765: Add concurrentAsyncExpand?
We have `Stream.asyncMap`, `concurrentAsyncMap`, `Stream.asyncExpand`... but no `concurrentAsyncExpand`.

```dart
StreamTransformer<S, T> concurrentAsyncExpand<S, T>(Stream<T> convert(S event))
```

Example:

```dart
Iterable<Future<int>> futures = [3, 2, 1].map((i) => Future.delayed(Duration(seconds: i), () => i));
Stream.fromIterable(futures)
  .transform(concurrentAsyncExpand((future) async* {
    yield 'awaiting future';
    var result = await future;
    yield 'future completed: $result';
  }))
  .forEach(print);
```

Output:
```
awaiting future
awaiting future
awaiting future
future completed: 1
future completed: 2
future completed: 3
```

natebosch:
Closes dart-lang/tools#1765

Added with the `Merge` extension, since the behavior is similar to
merging streams that are created lazily.

Adds a private `StreamTransformer` to merge a `Stream<Stream<T>>`. This
could be exposed separately as an extension on type.

In order to allow cancelling and then re-listening on broadcast streams,
an sub stream which is single subscription must be converted to
broadcast. This can be risky since subscriptions can leak and cause the
inner streams to never be canceled.

### DIFF
--- a/lib/src/merge.dart
+++ b/lib/src/merge.dart
@@ -54,6 +54,30 @@ extension Merge<T> on Stream<T> {
   /// events emitted by that stream before the result stream has a subscriber
   /// will be discarded.
   Stream<T> mergeAll(Iterable<Stream<T>> others) => transform(_Merge(others));
+
+  /// Like [asyncExpand] but the [convert] callback may be called for an element
+  /// before the Stream emitted by the previous element has closed.
+  ///
+  /// Events on the result stream will be emitted in the order they are emitted
+  /// by the sub streams, which may not match the order of the original stream.
+  ///
+  /// Errors from [convert], the source stream, or any of the sub streams are
+  /// forwarded to the result stream.
+  ///
+  /// The result stream will not close until the source stream closes and all
+  /// sub streams have closed.
+  ///
+  /// If the source stream is a broadcast stream the result will be as well,
+  /// regardless of the types of streams created by [convert]. In this case,
+  /// some care should be taken:
+  /// -  If [convert] returns a single subscription stream it may be listened to
+  /// and never canceled.
+  /// -  For any period of time where there are no listeners on the result
+  /// stream, any sub streams from previously emitted events will be ignored,
+  /// regardless of whether they emit further events after a listener is added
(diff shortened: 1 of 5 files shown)


lib/src/merge.dart
```diff
@@ -165,12 +165,12 @@ class _MergeExpanded<T> extends StreamTransformerBase<Stream<T>, T> {
       if (!streams.isBroadcast) {
         controller
           ..onPause = () {
-            for (var subscription in subscriptions) {
+            for (final subscription in subscriptions) {
```
grouma, line 168: Why this change?

grouma (approved):

lib/src/merge.dart
natebosch: It matches what I did in https://github.com/dart-lang/stream_transform/pull/92

As I have been rereading this code I found it easier to have a stronger signal for which variables were reassigned so I started replacing more `var` with `final`. It's less of a big deal in this small for loop, but I'm strongly considering turning on the lint to enforce it for this package.

PR #92, opened 2019-11-27
Refactor tracking of multiple stream subscriptions

natebosch:
Instead of using a variable with a bigger scope than the `onListen`
handler, and assigning to null in between calls to the callback, scope
the variable down. Remove the subscriptions when they are no longer
needed and use the size of the list as the subscription count rather
than tracking in a separate int variable.

Refactor a `..addAll` and a `.map` to use a list literal with a for-loop
element.

Make more local variables final.

### DIFF
--- a/lib/src/merge.dart
+++ b/lib/src/merge.dart
@@ -105,7 +105,6 @@ class _Merge<T> extends StreamTransformerBase<T, T> {
         final subscription =
             stream.listen(controller.add, onError: controller.addError);
         subscription.onDone(() {
-          assert(subscriptions.contains(subscription));
           subscriptions.remove(subscription);
           if (subscriptions.isEmpty) controller.close();
         });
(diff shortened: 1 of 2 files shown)


lib/src/merge.dart
```diff
@@ -63,47 +63,46 @@ class _Merge<T> extends StreamTransformerBase<T, T> {
...
-        activeStreamCount++;
-        return stream.listen(controller.add, onError: controller.addError,
-            onDone: () {
-          if (--activeStreamCount <= 0) controller.close();
+      final subscriptions = <StreamSubscription<T>>[];
+      for (final stream in allStreams) {
+        final subscription =
+            stream.listen(controller.add, onError: controller.addError);
+        subscription.onDone(() {
+          assert(subscriptions.contains(subscription));
```
grouma: Why is the assert necessary here and the similar one above?

grouma (approved):

lib/src/merge.dart
natebosch: Taking a look again, I think I had at one point during the refactor had a setup where the `subscriptions.remove` could happen in multiple places. Since this is now the only place that happens, and we know a stream will only end once, I agree that it's trivially safe. I'll remove them.

PR #95, opened 2019-12-03
Remove author from pubspec

natebosch:
Upon publish I was notified that this is no longer useful and can be
removed.


grouma:
I just ignore this lol...

grouma (approved):

PR #106, opened 2020-04-06
Add trailing call to throttle

maRci002:
Hello, I made some changes on throttle Stream to support trailing calls like in [underscorejs](https://underscorejs.org/docs/underscore.html#section-85) / [lodash](https://lodash.com/docs/4.17.15#throttle).

I encountered this problem while made some improvements on [flutter map](https://github.com/johnpryan/flutter_map/pull/572)  when I needed to detect tiles which should be downloaded while user is panning (basicly I didn't wanted to call _update tiles method every time during panning (performance reasons) so throttle was a good choice however I needed to make sure _update is recalled if a new value was added to stream during throttling).

```
source.throttle(Duration(seconds: 2));
source: 1-23----4---5-67|
result: 1-2-----4---5-6-|
```

```
source.throttle(Duration(seconds: 2), trailingCall: true);
source: 1-23----4---5-67|
result: 1-2-3---4---5-6-|
```

```
source.throttle(Duration(seconds: 2), trailingCall: true, trailing: true);
source: 1-23----4---5-67|
result: 1-2-3---4---5-6-7|
```

### DIFF
--- a/lib/src/rate_limit.dart
+++ b/lib/src/rate_limit.dart
@@ -84,17 +84,56 @@ extension RateLimit<T> on Stream<T> {
   ///
   /// Events emitted by the source stream within [duration] following an emitted
   /// event will be discarded. Errors are always forwarded immediately.
-  Stream<T> throttle(Duration duration) {
+  Stream<T> throttle(Duration duration,
+      {bool trailingCall = false, bool trailing = false}) {
     Timer timer;
+    T recentData;
+    var triggerTrailingCall = false;
+    var shouldClose = false;
 
-    return transform(fromHandlers(handleData: (data, sink) {
-      if (timer == null) {
-        sink.add(data);
-        timer = Timer(duration, () {
-          timer = null;
-        });
-      }
-    }));
+    void Function(T data, EventSink<T> sink) throttleHandler;
+    if (trailingCall) {
+      throttleHandler = (data, sink) {
+        recentData = data;
+        if (timer == null) {
+          sink.add(recentData);
+          if (shouldClose) {
+            sink.close();
+          } else {
+            timer = Timer(duration, () {
+              timer = null;
+              if (triggerTrailingCall) {
+                triggerTrailingCall = false;
+                throttleHandler(recentData, sink);
+              }
+            });
+          }
+        } else {
+          triggerTrailingCall = true;
+        }
+      };
+    } else {
+      throttleHandler = (data, sink) {
+        if (timer == null) {
+          sink.add(data);
+          timer = Timer(duration, () {
+            timer = null;
+          });
+        }
+      };
+    }
+
+    return transform(fromHandlers(
+      handleData: throttleHandler,
+      handleDone: (EventSink<T> sink) {
+        if (trailingCall && trailing) {
+          shouldClose = true;
+        } else {
+          timer?.cancel();
+          sink.close();
+        }
+      },
+    ));
   }
 
   /// Returns a Stream which only emits once per [duration], at the end of the
--- a/test/throttle_test.dart
+++ b/test/throttle_test.dart
@@ -73,6 +73,181 @@ void main() {
           });
         }
       });
+
+      group('throttle - trailingCall', () {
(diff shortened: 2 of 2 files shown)


maRci002:
@googlebot I signed it!

natebosch:
I don't understand your examples. Why are the `23` and `67` blocks different?

Can you explain in prose what these two new arguments do? How would you document them in the doc comment?

natebosch:
I don't understand, yet, whether this change is necessary or whether there are other operators that could give us the semantics you are hoping for, because I don't yet understand the semantics you are hoping for 😄 

See https://github.com/dart-lang/stream_transform/pull/103 and especially the doc comment changes there.

If we do decide we need a change along these lines it's likely that we'll want a separate `leading` and `trailing` argument to control the two ends of the duration.

I'd especially like to get a better understanding of the difference between `trailing` and `trailingCall`.

maRci002:
Here is some JS documentation for [trailingCall](https://gist.github.com/msssk/b720a8bddf2ba595347820ac387751ce).

`trailing` argument only takes effect if `trailingCall` is `true` and there is a pending trailing call so the stream will not closed before pending event is emitted.

maRci002:
`trailing` and `trailingCall` could be merged together but sometimes user don't want to get values when the source stream is closed.

maRci002:
I will add some documentation and merge `trailingCall` and `trailing` together (I just wanted to let user control pending `trailingCall` if the source stream is closed)

natebosch:
There are some similarities here to `audit`:

https://github.com/dart-lang/stream_transform/blob/ffdb9391dcccfe43ec4990cfdba0d171f211b48a/lib/src/rate_limit.dart#L121-L127

This looks about like a merge between the `throttle` and `audit` streams with the same duration.

One thing isn't clear from the linked docs - if an event is emitted soon after a "trailing" event was emitted should it be emitted or suppressed? The current implementation looks like it will emit it:

```
source:    1-2--3--|
throttled: 1----3--|
trailing:  1--2-3--|
```

Those events `2` and `3` are spaced closer together than the throttle duration, which is counter to the linked docs which state "only be triggered at most once during a given window of time".

The stated goal is:

> I needed to detect tiles which should be downloaded while user is panning (basicly I didn't wanted to call _update tiles method every time during panning (performance reasons) so throttle was a good choice however I needed to make sure _update is recalled if a new value was added to stream during throttling).

The way I think of this is:

> Don't give me more values until I'm ready for them, I'll be ready for more values some amount of time after I've handled a value.

I think the best operator to express this might be `asyncMapSample`. Instead of modeling the duration on the stream directly, it can be modeled by waiting to report that you have finished work for at least that duration.

```dart
var panEvents; // some stream that emits when the user pans
panEvents.asyncMapSample((_) async {
  _updateTiles();
  await Future.delayed(Duration(...));
}).drain();
```

Or you might even be able to relatively cheaply implement the operator you are hoping for with `asyncMapSample`. This would have slightly different (but I think preferred?) behavior where it wouldn't call for the "leading" event if it had too recently been called for the "trailing" event.

```dart
var panEvents; // some stream that emits when the user pans
var throttled = panEvents.asyncMapSample((event) async {
  await Future.delayed(Duration(...));
  return event;
});
throttled.listen((_) {
  _updateTiles();
});
```

What do you think?

maRci002:
Sorry for late response and thanks for suggestions @natebosch.

I think this version behaves like `audit`:
```dart
var panEvents; // some stream that emits when the user pans
var throttled = panEvents.asyncMapSample((event) async {
  await Future.delayed(Duration(...));
  return event;
});
throttled.listen((_) {
  _updateTiles();
});
```

This version works like charm:
```dart
var panEvents; // some stream that emits when the user pans
panEvents.asyncMapSample((_) async {
  _updateTiles();
  await Future.delayed(Duration(...));
}).drain();
```
The only problem with above version it emits trailing event even if source `stream` is closed, so if `_updateTiles` calls `setState` it has to check `mounted` property.

I think `asyncMapSample` resolves the problem thanks.

PR #109, opened 2020-05-04
Add sequentialAsyncExpand

creativecreatorormaybenot:
Similar to #90.

The reason this is useful is because sub streams generated by the `convert` callback for previous events from the source stream can be considered "outdated" in some use cases. In that case, still receiving events from previous sub streams (what `concurrentAsyncExpand` would do) is unwanted.

### Miscellaneous

I changed the homepage URL because the standard GitHub URL does not use a subdomain.  
However, this change is obviously unessential to this PR and can be removed if wanted.  
I did not want to open another PR for 1 loc changed.


creativecreatorormaybenot:
@grouma @natebosch Can you take a look? 🙂

natebosch:
Sorry I missed this!

This is the behavior of an existing transform [`switchMap`](https://pub.dev/documentation/stream_transform/latest/stream_transform/Switch/switchMap.html). If I swap them out these new tests pass.

We could maybe use this PR to instead add a reference to [`switchMap`] from the docs for [`concurrentAsyncExpand`], and vice-versa.

I'm not sure it would be worth the cost of a migration, but I do wonder if we should have chosen a different name for `switchMap`... that matches the rx libraries, but definitely would have been nice to use `expand` to match the existing Dart apis... `switchAsyncExpand` would be a more appropriate name I think.

creativecreatorormaybenot:
@natebosch How about creating an alias for it?

```dart
Stream<S> sequentialAsyncExpand<S>(Stream<S> Function(T) convert) => switchMap<S>(convert);
```

Although 🤔 Probably like this:

```dart
Stream<S> switchAsyncExpand<S>(Stream<S> Function(T) convert) => map(convert).switchLatest();
```

+ ***deprecate*** `switchMap`.

Do you want me to make the documentation change?

natebosch:
If we decide to rename it we'd definitely create an alias first and `@Deprecate()` the original. I'd prefer not to have 2 names for it in the long term.

> Do you want me to make the documentation change?

If you'd like to push a commit dropping the new API and keeping the "see also" doc comment change pointing to `switchMap` (we'll need to add an import for the doc reference link to work, I think that should be fine to do) we can get that merged in. We'll hold off for now on the alias and deprecation. I'll sync with some other folks on the team next week regarding the name change.

creativecreatorormaybenot:
@natebosch Done.

natebosch (approved):

natebosch:
Thanks!

PR #119, opened 2020-12-01
Migrate to null safety

natebosch:
- Bump SDK constraint.
- Bump to null safe dependencies.
- Migrate code, examples, and tests.
- Add an opt-out test for testing the behavior of transforming streams
  with the `subscription.cancel()` call returns null despite a
  statically non-nullable return. This requires an unusual pattern in
  the code.

### DIFF
--- a/lib/src/aggregate_sample.dart
+++ b/lib/src/aggregate_sample.dart
@@ -44,7 +44,7 @@ class AggregateSample<S, T> extends StreamTransformerBase<S, T> {
       if (!waitingForTrigger) emit();
 
       if (isTriggerDone) {
-        valueSub?.cancel();
+        valueSub!.cancel();
         controller.close();
       }
     }
@@ -63,7 +63,7 @@ class AggregateSample<S, T> extends StreamTransformerBase<S, T> {
       if (currentResults != null) emit();
 
       if (isValueDone) {
-        triggerSub?.cancel();
+        triggerSub!.cancel();
         controller.close();
       }
     }
--- a/lib/src/aggregate_sample.dart
+++ b/lib/src/aggregate_sample.dart
@@ -15,7 +15,7 @@ import 'dart:async';
 /// the output.
 class AggregateSample<S, T> extends StreamTransformerBase<S, T> {
   final Stream<void> _trigger;
-  final T Function(S, T) _aggregate;
+  final T Function(S, T?) _aggregate;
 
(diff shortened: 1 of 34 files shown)


lib/src/aggregate_sample.dart
```diff
@@ -44,7 +44,7 @@ class AggregateSample<S, T> extends StreamTransformerBase<S, T> {
       if (!waitingForTrigger) emit();
 
       if (isTriggerDone) {
-        valueSub.cancel();
+        valueSub?.cancel();
```
jakemac53: It seems a bit weird to change this to `?` - it was previously the equivalent of a `!`. Maybe just a more defensive change here though? (same for `triggerSub` below)

lib/src/combine_latest.dart
```diff
@@ -249,11 +253,10 @@ class _CombineLatestAll<T> extends StreamTransformerBase<T, List<T>> {
           };
       }
       controller.onCancel = () {
-        var cancels = subscriptions
-            .map((s) => s.cancel())
-            .where((f) => f != null)
-            .toList();
-        if (cancels.isEmpty) return null;
```
jakemac53, line 256: This was retained for other code paths, might as well retain it here

lib/src/switch.dart
```diff
@@ -96,8 +96,10 @@ class _SwitchTransformer<T> extends StreamTransformerBase<Stream<T>, T> {
       controller.onCancel = () {
         var cancels = [
           if (!outerStreamDone) outerSubscription.cancel(),
-          if (innerSubscription != null) innerSubscription.cancel(),
-        ].where((f) => f != null);
+          if (innerSubscription != null) innerSubscription!.cancel(),
```
jakemac53, line 99: Since we remove nulls you could do this, 🤷 
```suggestion
         innerSubscription?.cancel(),
```

jakemac53 (approved):

lib/src/aggregate_sample.dart
natebosch: Good point. I'll switch it back.

lib/src/combine_latest.dart
natebosch: Done.

lib/src/switch.dart
natebosch: I decided to keep this as is so that for Dart 2 we remove the 2 lines below without needing to change this.

PR #120, opened 2020-12-04
Refactor to remove StreamTransformer instantiation

natebosch:
Directly implement the behavior from `bind` which is the `Stream` to
`Stream` implementation in the extension methods rather than in separate
`StreamTransformer` classes. This pattern has fewer potential issues
around generics since it's fully based on function calls which are
type checked appropriately unlike the `StreamTransformer` interface
which suffers from the incorrect covariant subtype relationships. It
also may avoid some unnecessary allocations as improve stack traces.

For the transforms that were changed, aim for better consistency around
generic type arguments. `T` is now more consistently the source stream
type, while `S` is the output stream type.

### DIFF
--- a/lib/src/from_handlers.dart
+++ b/lib/src/from_handlers.dart
@@ -5,6 +5,11 @@
 import 'dart:async';
 
 extension TransformByHandlers<S> on Stream<S> {
+  /// Transform a stream by callbacks.
+  ///
+  /// This is similar to `transform(StreamTransformer.fromHandler(...))` except
+  /// that the handlers are called once per event rather than called for the
+  /// same event for each listener on a broadcast stream.
   Stream<T> transformByHandlers<T>(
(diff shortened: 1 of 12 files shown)


lib/src/from_handlers.dart
```diff
@@ -4,61 +4,31 @@
...
-  }
-
-  static void _defaultHandleDone<T>(EventSink<T> sink) {
-    sink.close();
-  }
-
-  @override
-  Stream<T> bind(Stream<S> values) {
-    var controller = values.isBroadcast
+extension TransformByHandlers<S> on Stream<S> {
```
grouma, line 7: [nit] Maybe a doc comment here?

grouma (approved):

lib/src/from_handlers.dart
natebosch: Done

PR #123, opened 2021-01-11
Add trailing argument to throttle

### LINKED ISSUE #122: Add a trailing option to throttle
I discussed the need for a rate limiting method with @natebosch that has the following properties:

- Emit the first event at the beginning of the period
- Emit the last event at the end of the period

Given a period of 500ms (one `-` represents 100ms):
```
Source:  A-B----C-D-------E--F
Periods: |----|----|----  |----|----
Result:  A----B----D------E----F
```

This is similar to both `audit` and `throttle` with some key differences:
- Unlike `audit`, this doesn't delay the initial event (or the initial event after a period has elapsed since the last event)

    ```
    Source:  A-B----C-D-------E--F
    Periods: ----|  ----|     ----|
    Result:  ----B------D---------F
    ```
- Unlike `throttle`, this won't "miss" the last event in a period

    ```
    Source:  A-B----C-D-------E--F
    Periods: |----  |----     |----
    Result:  A------C---------E
    ```

Nate pointed out that this behavior could be implemented today using `asyncMapSample`:

```
Stream<T> firstAndLastInDuration(Stream<T> stream, Duration duration) {
  var controller = StreamController<T>(sync: true);
  stream.asyncMapSample((event) async {
    controller.add(event);
    await Future.delayed(duration);
  }).drain();
  return controller.stream;
}
```

Alternatively, the same behavior could be provided by adding a `trailing` argument to `throttle` which would make it also emit the last event in the period (and reset the timer). This would give it the exact behavior described above:

```
Source:  A-B----C-D-------E--F
Periods: |----|----|----  |----|----
Result:  A----B----D------E----F
```

### LINKED ISSUE #1776: Add a trailing option to throttle
I discussed the need for a rate limiting method with @natebosch that has the following properties:

- Emit the first event at the beginning of the period
- Emit the last event at the end of the period

Given a period of 500ms (one `-` represents 100ms):
```
Source:  A-B----C-D-------E--F
Periods: |----|----|----  |----|----
Result:  A----B----D------E----F
```

This is similar to both `audit` and `throttle` with some key differences:
- Unlike `audit`, this doesn't delay the initial event (or the initial event after a period has elapsed since the last event)

    ```
    Source:  A-B----C-D-------E--F
    Periods: ----|  ----|     ----|
    Result:  ----B------D---------F
    ```
- Unlike `throttle`, this won't "miss" the last event in a period

    ```
    Source:  A-B----C-D-------E--F
    Periods: |----  |----     |----
    Result:  A------C---------E
    ```

Nate pointed out that this behavior could be implemented today using `asyncMapSample`:

```
Stream<T> firstAndLastInDuration(Stream<T> stream, Duration duration) {
  var controller = StreamController<T>(sync: true);
  stream.asyncMapSample((event) async {
    controller.add(event);
    await Future.delayed(duration);
  }).drain();
  return controller.stream;
}
```

Alternatively, the same behavior could be provided by adding a `trailing` argument to `throttle` which would make it also emit the last event in the period (and reset the timer). This would give it the exact behavior described above:

```
Source:  A-B----C-D-------E--F
Periods: |----|----|----  |----|----
Result:  A----B----D------E----F
```

natebosch:
Closes dart-lang/tools#1776

The existing rate limit utilities don't cover the case of wanting to
update a UI with some changing data with a limited frequency. In this
case the goal is to get the "latest" event with minimal lag time, while
maintaining at least a given duration in between events. Add a
`trailing` argument to `throttle` for an operator which will emit
immediately when possible (unlike `audit`), and does not get starved by a
long series of frequent events (unlike `debounce`).

Expand the doc with the same timeline diagrams as used for other
operators.

Split the implementation into two methods to avoid extra conditional
checks and unused variables for the common case of `trailing: false`.
Add tests for the behavior of emitting the trailing event, suppressing
intermediate events, and keeping the stream open until the final
trailing even is emitted.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 - Migrate to null safety.
 - Improve tests of `switchMap` and improve documentation with links and
   clarification.
+- Add `trailing` argument to `throttle`.
 
 ## 1.2.0
 
--- a/lib/src/rate_limit.dart
+++ b/lib/src/rate_limit.dart
@@ -73,7 +73,7 @@ extension RateLimit<T> on Stream<T> {
   /// debounce period before closing. If there are no pending debounced events
   /// when the source stream closes the returned stream will close immediately.
   ///
-  /// To keep only the most recent event during the debounce perios see
+  /// To keep only the most recent event during the debounce period see
   /// [debounce].
   Stream<List<T>> debounceBuffer(Duration duration) =>
       _debounceAggregate(duration, _collect, leading: false, trailing: true);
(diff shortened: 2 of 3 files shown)


grouma (approved):
LGTM. Maybe call out in the documentation that the last pending event will be emitted if the sink is closed and `trailing` is `true`.

natebosch:
> Maybe call out in the documentation that the last pending event will be emitted if the sink is closed and `trailing` is `true`.

I have the sentence "If the source stream closes the result stream will wait to emit a pending event before closing." in the doc to cover that.

PR #128, opened 2021-02-08
Publish stable version of stream_transform

kevmoo:


kevmoo:
CC @lukaszkolodziejczyk

CHANGELOG.md
```diff
@@ -1,4 +1,4 @@
-## 2.0.0-nullsafety.0
+## 2.0.0
```
natebosch, line 1: We have been keeping the `-nullsafety.0` changelog entries in place and adding new ones for the stable release.

Should we be changing our approach? cc @jakemac53

jakemac53: I don't know... I can see an argument either way.

In the case of breaking changes, they can be a bit buried with the current approach.

kevmoo: My thoughts exactly. The null safe versions are blips in history. Noise for the vast majority of users.

natebosch (approved):

PR #135, opened 2021-07-26

natebosch:
cleanup is done.

### DIFF
--- a/lib/src/switch.dart
+++ b/lib/src/switch.dart
@@ -62,8 +62,8 @@ extension SwitchLatest<T> on Stream<Stream<T>> {
       var outerStreamDone = false;
 
       final outerSubscription = listen(
-          (innerStream) {
-            innerSubscription?.cancel();
+          (innerStream) async {
+            await innerSubscription?.cancel();
             innerSubscription = innerStream.listen(controller.add,
                 onError: controller.addError, onDone: () {
               innerSubscription = null;
--- a/test/switch_test.dart
+++ b/test/switch_test.dart
@@ -113,25 +113,24 @@ void main() {
           expect(firstCanceled, true);
         });
 
-        test('waits for cancel before listening to subsequent stream',
-            () async {
-          var cancelWork = Completer<void>();
-          first.onCancel = () => cancelWork.future;
-          outer.add(first.stream);
-          await Future(() {});
+        if (innerType != 'broadcast') {
+          test('waits for cancel before listening to subsequent stream',
+              () async {
+            var cancelWork = Completer<void>();
+            first.onCancel = () => cancelWork.future;
+            outer.add(first.stream);
+            await Future(() {});
 
-          var cancelDone = false;
-          second.onListen = expectAsync0(() {
-            expect(cancelDone, true);
+            var cancelDone = false;
+            second.onListen = expectAsync0(() {
+              expect(cancelDone, true);
+            });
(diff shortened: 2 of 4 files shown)


lib/src/switch.dart
```diff
@@ -62,8 +62,8 @@ extension SwitchLatest<T> on Stream<Stream<T>> {
       var outerStreamDone = false;
 
       final outerSubscription = listen(
-          (innerStream) {
-            innerSubscription?.cancel();
+          (innerStream) async {
+            await innerSubscription?.cancel();
```
lrhn, line 66: Not sure it's worth waiting for the `cancel` to complete if you don't handle an error anyway.

If `cancel` returns an error future, then that error becomes an uncaught error. The `listen` callbacks are not supposed to throw, and if they do, the error just becomes uncaught.
The `await` means that the following code doesn't run in that case, which it did before.

If the `cancel` completes normally, you wait for it, and that *might* be reasonable, but also might just be a delay.

Since you don't pause the outer stream while waiting, this code may be reentered before it's done, if another stream arrives while waiting for the `cancel`. It'll then try to cancel the same `innerSubscription` twice, then eventually listen on two (or more) streams, and not be able to cancel all of them.

Maybe something like:
```dart
try {
  outerSubscription.pause();
  await innerSubscription?.cancel();
} finally {
  the rest
  outerSubscription.resume();
}
```
(or use `.whenComplete` directly instead of `async` + `try`/`finally`).

Or, alternatively, if you don't want to pause the outer stream, you need some state management. If a new stream arrives while waiting for an old stream to cancel, you just override the "pending stream" with the new stream, and never listen to to the stream that got superseeded before it was started (unless you actually need/want to wait `listen`+`cancel` the stream to not leave the creator dangling - shouldn't be necessary, but sometimes is).

That's a larger rewrite (but not unexpected for something involving multiple streams or other async operations).

test/switch_test.dart
```diff
@@ -113,6 +113,25 @@ void main() {
...
+            var cancelWork = Completer<void>();
+            first.onCancel = () => cancelWork.future;
+            outer.add(first.stream);
+            await Future(() {});
+
+            var cancelDone = false;
+            second.onListen = expectAsync0(() {
+              expect(cancelDone, true);
+            });
+            outer.add(second.stream);
```
lrhn, line 128: Try adding a third stream here as well, and ensure that the behavior is still desirable (and predictable, so document what happens).

lib/src/switch.dart
lrhn: Also, if the returned stream is paused during an inner-stream cancel, you still listen to the next inner-stream afterwards, and it isn't paused.

lrhn:
I tried making my own version, addressing all the comments here. https://github.com/dart-lang/stream_transform/pull/136

It still suffers from the problem that if more than one stream arrives during a cancel, the non-latest ones are completely ignored.
We might want to actually listen-to-and-cancel each of them in turn, which means needing more state (possibly even a queue of pending streams).
The alternative is to pause the outer stream while cancelling. That feels wrong for a stream which is all about the *latest* event, but if we need to address each stream along the way, at least just to listen+cancel, then buffering in the subscription, which it already supports, beats introducing a new buffer.

So, I did another which uses pause: https://github.com/dart-lang/stream_transform/pull/137


It's amazing the state-space explosion a single delay in a seemingly simple algorithm can cause!

PR #136, opened 2021-07-27

lrhn:
cleanup is done.


natebosch:
Closing in favor of #137

PR #137, opened 2021-07-27

lrhn:
cleanup is done.

### DIFF
--- a/lib/src/switch.dart
+++ b/lib/src/switch.dart
@@ -47,11 +47,12 @@ extension Switch<T> on Stream<T> {
 extension SwitchLatest<T> on Stream<Stream<T>> {
   /// Emits values from the most recently emitted Stream.
   ///
-  /// When the source emits a stream the output will switch to emitting events
+  /// When the source emits a stream, the output will switch to emitting events
   /// from that stream.
   ///
-  /// If the source stream is a broadcast stream, the result stream will be as
-  /// well, regardless of the types of streams emitted.
+  /// Whether the source stream is a single-subscription stream or a
+  /// broadcast stream, the result stream will be the same kind of stream,
+  /// regardless of the types of streams emitted.
   Stream<T> switchLatest() {
     var controller = isBroadcast
         ? StreamController<T>.broadcast(sync: true)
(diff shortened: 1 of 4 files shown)


lrhn:
This is an alternative version using `pause` on the outer stream while cancelling. The logic is simpler, and all streams are listened to (even if they're then almost immediately cancelled).

natebosch (approved):

lib/src/switch.dart
```diff
@@ -61,20 +61,38 @@ extension SwitchLatest<T> on Stream<Stream<T>> {
...
+          listen(null, onError: controller.addError, onDone: () {
+        outerStreamDone = true;
+        if (innerSubscription == null) controller.close();
+      });
+      outerSubscription.onData((innerStream) {
+        var currentSubscription = innerSubscription;
+        if (currentSubscription != null) {
+          innerSubscription = null;
+          outerSubscription.pause();
+          currentSubscription.cancel().whenComplete(() {
```
natebosch: Thinking more about it I'm leaning towards forwarding errors to the result stream.

test/switch_test.dart
```diff
@@ -49,16 +51,12 @@ void main() {
         test('forwards events', () async {
           outer.add(first.stream);
           await Future(() {});
-          first
-            ..add(1)
-            ..add(2);
+          first..add(1)..add(2);
```
natebosch: It looks like your SDK has an older version of dart format. If you update to the latest dev SDK the formatting should match CI.

natebosch:
@googlebot I consent

lib/src/switch.dart
lrhn: That should be possible. I'd do something like:
```dart
Future<void>.sync(currentSubscription.cancel).catchError(controller.addError).whenComplete(() {
  ...
});
```
That should also catch synchronous errors from `cancel` (just to be paranoid).
I *think* it's safe to use `controller.addError` unconditionally. There is no chance that it has been closed if we paused `outerStream` when we received the previous stream event which caused the current cancel.

lrhn:
PTAL. Added forwarding of errors.

natebosch:
I'm working with some teams to resolve some internal failures before we land this.

natebosch (approved):

PR #139, opened 2021-11-30
Refactor a callback to async/await

natebosch:
Use a try/finally block over using `whenComplete`. This also allows
combining the async and sync error handling with a single `catch`
instead of using `catch` for sync errors and `catchError` for async
errors.

Switch the early return to the shorter case of immediately listening on
the first stream, and leaving the more complex logic (and more common
case) with one less level of nesting.

This does move the call to `outerSubscription.pause()` to come before
the all to `currentSubscription.cancel()`, but this ordering should not
be meaningful even though it can be observed.

Fix a TODO about using `Future.ignore()`.

### DIFF
--- a/lib/src/switch.dart
+++ b/lib/src/switch.dart
@@ -81,36 +81,31 @@ extension SwitchLatest<T> on Stream<Stream<T>> {
         outerStreamDone = true;
         if (innerSubscription == null) controller.close();
       });
-      outerSubscription.onData((innerStream) {
+      outerSubscription.onData((innerStream) async {
         var currentSubscription = innerSubscription;
-        if (currentSubscription != null) {
-          innerSubscription = null;
-          try {
-            currentSubscription.cancel().catchError(addError).whenComplete(() {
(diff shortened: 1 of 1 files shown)


natebosch:
Friendly ping @lrhn

lib/src/switch.dart
```diff
@@ -81,36 +81,31 @@ extension SwitchLatest<T> on Stream<Stream<T>> {
...
+        } catch (error, stack) {
+          controller.addError(error, stack);
+        } finally {
+          if (!isBroadcast && !controller.hasListener) {
+            // Result single-subscription stream subscription was cancelled
+            // while waiting for previous innerStream cancel.
+            //
+            // Ensure that the last received stream is also listened to and
+            // cancelled, then do nothing further.
+            innerStream.listen(null).cancel().ignore();
```
lrhn, line 103: Exactly the intended kind of use-case for `ignore`. Glad to see it used!

lrhn (approved):
LGTM!

PR #147, opened 2022-05-20
Implement `sample`, add `longPoll` to `buffer`

### LINKED ISSUE #145: Consider adding an argument to `buffer` to match ReactiveX behavior
While I was trying to more deeply understand the reactive docs I found an interesting behavior difference between our `buffer` and reactive [`buffer(bufferClosingSelector)`](https://reactivex.io/documentation/operators/buffer.html) which I suspect makes them fit fairly different use cases.

In Reactive, if you send a "close" selector when there are no events buffered you get an empty list immediately. In `stream_transform` if you send a trigger when there are no events buffered you get set up to be immediately notified of the next even in a single item list. This is a bit like a "long polling" behavior for buffering, vs the strict immediate polling from reactive.

Typically I don't like to add speculative generality, however in this case it might be worth adding the support eagerly since the additional generality may actually make it easier to document and compare to Rx.

### LINKED ISSUE #1781: Consider adding an argument to `buffer` to match ReactiveX behavior
While I was trying to more deeply understand the reactive docs I found an interesting behavior difference between our `buffer` and reactive [`buffer(bufferClosingSelector)`](https://reactivex.io/documentation/operators/buffer.html) which I suspect makes them fit fairly different use cases.

In Reactive, if you send a "close" selector when there are no events buffered you get an empty list immediately. In `stream_transform` if you send a trigger when there are no events buffered you get set up to be immediately notified of the next even in a single item list. This is a bit like a "long polling" behavior for buffering, vs the strict immediate polling from reactive.

Typically I don't like to add speculative generality, however in this case it might be worth adding the support eagerly since the additional generality may actually make it easier to document and compare to Rx.

natebosch:
Closes dart-lang/tools#1781

Add `longPoll` support to `aggregateSample`. This method is a package
private implementation detail.
- Refactor to named arguments for clarity at the calling sites.
- Add `longPoll` boolean argument and conditional behavior in each
  stream callback.
- Add `onEmpty` callback to let the caller decide whether to emit some
  default (empty list in the `buffer` case) or not emit at all (in the
  `sample` case) for triggers which do no set up a long poll and have
  no prior aggregated events.
- Switch `waitingForTrigger` to the inverted condition `activeLongPoll`
  since its a closer match to how the behavior is described in the docs.

Add `longPoll` argument to `buffer` and update the doc to describe the
conditional behavior. Add a section in the docs discussing the end of
stream behavior.

Add a `sample` method which is similar to `buffer`. Cross reference each
method in the docs.

Update the comparison table with Rx. Show that `buffer` is equivalent to
`buffer(longPoll: false)`, that `sample` is `sampel(longPoll: false)`. I
cannot find an equivalent Rx method to the `longPoll: true` behavior.

### DIFF
--- a/lib/src/aggregate_sample.dart
+++ b/lib/src/aggregate_sample.dart
@@ -5,41 +5,54 @@
 import 'dart:async';
 
 extension AggregateSample<T> on Stream<T> {
-  /// Aggregates values and emits when it sees a value on [trigger].
+  /// Aggregates values and polls the result when there is an event emitted by
+  /// [trigger].
   ///
-  /// If there are no pending values when [trigger] emits, the next value on the
-  /// source Stream will be passed to [aggregate] and emitted on the result
-  /// stream immediately. Otherwise, the pending values are released when
-  /// [trigger] emits.
+  /// If [longPoll] is `false`, then an event on [trigger] when there is no
+  /// pending aggregated result will immediately invoke [onEmpty] with a sink
+  /// for the result stream.
+  /// If [longPoll] is `true` (the default), then an event on [trigger] when
+  /// there are is no pending aggregated result will cause the next source event
+  /// to immediately flow to the result stream as the aggregate conversion of
+  /// the single element. If [longPoll] is `true` [onEmpty] is never used.
+  ///
+  /// The result stream will close as soon as there is a guarantee it will not
+  /// emit any more events. There will not be any more events emitted if:
+  /// - [trigger] is closed and there is no waiting long poll.
+  /// - Or, the source stream is closed and there are no buffered events.
   ///
   /// Errors from the source stream or the trigger are immediately forwarded to
   /// the output.
   Stream<S> aggregateSample<S>(
-      Stream<void> trigger, S Function(T, S?) aggregate) {
+      {required Stream<void> trigger,
+      required S Function(T, S?) aggregate,
+      required bool longPoll,
+      required void Function(Sink<S>) onEmpty}) {
     var controller = isBroadcast
         ? StreamController<S>.broadcast(sync: true)
         : StreamController<S>(sync: true);
 
     S? currentResults;
     var hasCurrentResults = false;
-    var waitingForTrigger = true;
+    var activeLongPoll = false;
     var isTriggerDone = false;
     var isValueDone = false;
     StreamSubscription<T>? valueSub;
     StreamSubscription<void>? triggerSub;
 
-    void emit() {
-      controller.add(currentResults as S);
+    void emit(S results) {
+      controller.add(results);
       currentResults = null;
       hasCurrentResults = false;
-      waitingForTrigger = true;
+      activeLongPoll = false;
     }
 
     void onValue(T value) {
       currentResults = aggregate(value, currentResults);
       hasCurrentResults = true;
+      if (!longPoll) return;
 
-      if (!waitingForTrigger) emit();
+      if (activeLongPoll) emit(currentResults as S);
 
       if (isTriggerDone) {
         valueSub!.cancel();
@@ -56,9 +69,13 @@ extension AggregateSample<T> on Stream<T> {
     }
 
     void onTrigger(_) {
-      waitingForTrigger = false;
-
-      if (hasCurrentResults) emit();
+      if (hasCurrentResults) {
+        emit(currentResults as S);
+      } else if (longPoll) {
+        activeLongPoll = true;
+      } else {
+        onEmpty(controller);
+      }
 
       if (isValueDone) {
         triggerSub!.cancel();
@@ -68,7 +85,7 @@ extension AggregateSample<T> on Stream<T> {
 
     void onTriggerDone() {
       isTriggerDone = true;
-      if (waitingForTrigger) {
+      if (!activeLongPoll) {
         valueSub?.cancel();
         controller.close();
       }
--- a/lib/src/aggregate_sample.dart
+++ b/lib/src/aggregate_sample.dart
@@ -5,22 +5,33 @@
 import 'dart:async';
 
 extension AggregateSample<T> on Stream<T> {
-  /// Aggregates values and polls the result when there is an event emitted by
-  /// [trigger].
+  /// Computes a value based on sequences of events, then emits that value when
+  /// [trigger] emits an event.
   ///
-  /// If [longPoll] is `false`, then an event on [trigger] when there is no
-  /// pending aggregated result will immediately invoke [onEmpty] with a sink
-  /// for the result stream.
-  /// If [longPoll] is `true` (the default), then an event on [trigger] when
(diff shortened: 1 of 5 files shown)


lib/src/aggregate_sample.dart
```diff
@@ -5,41 +5,54 @@
 import 'dart:async';
 
 extension AggregateSample<T> on Stream<T> {
-  /// Aggregates values and emits when it sees a value on [trigger].
+  /// Aggregates values and polls the result when there is an event emitted by
+  /// [trigger].
```
lrhn: This first line doesn't tell me what the method does. Can it be more precise?

If the first line is all you read, you should still have a clue whether the method is right for the problem you have. I have no clue after reading this. I don't know which values we are talking about (assumes it's value events of `this`, but not sure), don't know what "polls" mean, and don't know what "result" refers to.

So, after reading the code and variables, my guess is:

For each event of [this], we call [aggregate] with either the previous result of calling [aggregate], or `null` if none, and the event. When the [trigger] stream emits an event, the returned stream emits the most recently computed result. (I don't know if that resets the aggregate to `null`, but the `longPoll` parameter suggests that it does reset).

I think it's a slightly weird functionalty. The `trigger` stream makes it look like it's pull based and should really be an iterable instead. (With `longPoll` it would have to return a future, though).

Since `trigger` is a stream, that it has no values is also surprising. I'd consider allowing it to have a value of type `R`, and emitting the result of a `V Function(R, S)` function. A trigger stream is not something I've seen before. 

This also sounds like something you can write using a general parallel processing operation: Take two streams, for each even, call a function, maybe differ for the streams, with an intermediate value and the new event, to create a new intermediate event. Then you just let one function "aggregate" the results and the other send the result to a stream controller.

How about:
```dart
/// Computes a value based on sequences of events, then emits that value when [trigger] emits an event.
///
/// Every time [this] stream emits an event, an intermediate value is created by combining the new event
/// with the previous intermediate value, or with `null` if there is no previous value, using the [aggregate]
/// function.
/// When [trigger] emits value, the returned stream emits the current intermediate value and clears it.
/// If [longPoll] is `false`, if there is no intermediate value when [trigger] emits a value, the [onEmpty]
/// function is called with a [Sink] which can add events to the returned stream.
/// If [longPoll] is `true`, and there is no intermediate value, then the *next* event from [this] stream
/// is immediately put through [aggregate] and then emitted on the returned stream.
/// In that case, [onEmpty] is never used. (-- so why is it not optional? --)
/// (-- what if more than one trigger happens during a long poll? --)
///
/// ...
```
I'd make `onEmpty` optional, and emit an error (`StateError("no value")`) if it would have been called.
I'd make `longPoll` optional and default to `false`.

I'd make trigger and aggregate positional parameters.

lib/src/rate_limit.dart
```diff
@@ -224,15 +224,60 @@ extension RateLimit<T> on Stream<T> {
   /// Returns a Stream  which collects values and emits when it sees a value on
```
lrhn: Stream -> \[Stream\]  (or just "stream" as a word instead of a source name).

Double space after "Stream  "

Should use the word "buffer" in the first line.

lib/src/rate_limit.dart
```diff
@@ -224,15 +224,60 @@ extension RateLimit<T> on Stream<T> {
...
+  /// - [sample] which use a [trigger] stream in the same way, but keeps only
+  /// the most recent source event.
+  Stream<List<T>> buffer(Stream<void> trigger, {bool longPoll = true}) =>
+      aggregateSample(
+          trigger: trigger,
+          aggregate: _collect,
+          longPoll: longPoll,
+          onEmpty: _empty);
+
+  /// Returns a Stream which emits the most recent new value from the source
```
lrhn: Returns -> Creates

\[Stream\]

lib/src/aggregate_sample.dart
```diff
@@ -5,41 +5,54 @@
 import 'dart:async';
 
 extension AggregateSample<T> on Stream<T> {
```
lrhn, line 7: Is this method not exported by the `async.dart` library?

If it's internal, then most of the comments here can be ignored (like all constructor arguments being required is fine, I'd personally make them positional.)

lib/src/aggregate_sample.dart
```diff
@@ -5,41 +5,54 @@
...
     StreamSubscription<void>? triggerSub;
 
-    void emit() {
-      controller.add(currentResults as S);
+    void emit(S results) {
+      controller.add(results);
       currentResults = null;
       hasCurrentResults = false;
-      waitingForTrigger = true;
+      activeLongPoll = false;
```
lrhn: Could move `activeLongPoll = false` into `onValue`, just before calling `emit`.
I looked for it there, and it won't be true if this function is called from `onTrigger`.

lib/src/aggregate_sample.dart
```diff
@@ -5,41 +5,54 @@
...
+    var activeLongPoll = false;
     var isTriggerDone = false;
     var isValueDone = false;
     StreamSubscription<T>? valueSub;
     StreamSubscription<void>? triggerSub;
 
-    void emit() {
-      controller.add(currentResults as S);
+    void emit(S results) {
+      controller.add(results);
```
lrhn: The controller is sync, so I'd clear the state before calling `add`.
```dart
  var current = currentResults;
  currentResults = null;
  hasCurrentResults = false;
  activeLongPoll = false; // if not moved.
  controller.add(current);
```

lrhn (approved):

lib/src/aggregate_sample.dart
natebosch: Yes this is internal. (I'd mark it with `@internal` but we don't currently have any non-dev dependencies and it doesn't seem worth adding `meta`)


I'll keep the named arguments for now. As I was trying out different approaches to this implementation I found it easier to work with the changing code using named arguments. If we decide to make this API, or something similar, public then we can switch to positional.

lib/src/aggregate_sample.dart
natebosch: I took most of your suggested phrasing. Let me know if you'd like to see further changes to any of the docs in a followup. I'm happy to bikeshed a bit since it impacts the user facing docs for `sample` and `buffer`

lib/src/aggregate_sample.dart
natebosch: Done.

lib/src/aggregate_sample.dart
natebosch: Done.

lib/src/rate_limit.dart
natebosch: Done

lib/src/rate_limit.dart
natebosch: Done

PR #148, opened 2022-05-24
Relisten to ongoing streams in concurrentAsyncExpand

natebosch:
Remove an unnecessary `asBroadcastStream` call. This had always been
unnecessary since each inner stream was only listened to once when it is
seen for the first time. Now some streams are listened to more than
once, but only the broadcast streams.

Keep the `Stream` instance along with the subscription to allow
re-listening.

Only cancel subscriptions inner broadcast streams. Subscriptions on
inner single-subscriber streams will remain open, but the events the add
to the controller will be ignored while the result stream has no
listeners. When listeners are added again, resubscribe to the ongoing
inner broadcast streams.


natebosch:
I also had another version of this where I don't `.cancel()` the inner subscriptions at all. It's a little simpler, but I think it's worth the slight extra complexity here to allow the `onCancel` to work correctly for broadcast streams.

lib/src/async_expand.dart
```diff
@@ -45,41 +44,52 @@ extension AsyncExpand<T> on Stream<T> {
         ? StreamController<S>.broadcast(sync: true)
         : StreamController<S>(sync: true);
 
+    var outerStreamDone = false;
+    final subscriptions = <Stream<S>, StreamSubscription<S>>{};
```
lrhn: So, you remember the stream because when the outer stream is a broadcast stream and it stops having listeners, you unsubscribe from inner broadcast streams, then resubscribe when the outer stream gets listeners again (if it's broadcast, otherwise it should just stop).
See above for comment on that.

It feels wrong/unnecessary to remember the stream for non-broadcast inner streams, though.
And the map is treated more like a set of pairs than used as a map. (Which is good, because we have not defined equality on streams, so you shouldn't assume they're using identical. I'd prefer to use `LinkedHashMap.identity` for this instead, to ensure you use identity on the keys.)

Consider instead creating some helper class to contain subscription and stream for broadcast streams, and just the subscription for non-broadcast streams, and have a set of instances of that instead.

lib/src/async_expand.dart
```diff
@@ -45,41 +44,52 @@ extension AsyncExpand<T> on Stream<T> {
...
+            for (final subscription in subscriptions.values) {
               subscription.resume();
             }
           };
       }
       controller.onCancel = () {
-        if (subscriptions.isEmpty) return null;
-        var cancels = [for (var s in subscriptions) s.cancel()]
+        final toCancel = {
+          if (!outerStreamDone) outerSubscription,
```
lrhn: Consider calling `cancel` on the subscriptions immediately in this literal, instead of first building a list of subscriptions, then building a list of calling `cancel` on each of them.
```dart
var cancels = [for (var entry in subsciprtions.entries)
  if (!isBroadcast || entry.key.isBroadcast) entry.value.cancel()
];
```

lib/src/async_expand.dart
```diff
@@ -32,9 +32,8 @@ extension AsyncExpand<T> on Stream<T> {
   /// -  If [convert] returns a single subscription stream it may be listened to
   /// and never canceled.
   /// -  For any period of time where there are no listeners on the result
-  /// stream, any sub streams from previously emitted events will be ignored,
-  /// regardless of whether they emit further events after a listener is added
-  /// back.
+  /// stream, any event on the source stream is ignored and will not be passed
+  /// to [convert].
```
lrhn: Is that desirable behavior?

That is, if an event would come on the (broadcast) source stream while there is no listeners, that event can still turn into a stream which will emit events later, when there are listeners.
By dropping the source stream event, you are dropping any number of future events, also some which would appear while there are listeners. That breaks the broadcast-stream contract, which is that it *shouldn't* matter whether you unsubscribe and resubscribe for a while, or just ignore events for the same while. 

So, I'd keep listening on the source stream and create new streams. You can delay subscribing to those streams  until the outer stream gets a listener again, if they are broadcast streams. I wouldn't delay subscribing to non-broadcast streams.

lrhn (approved):

lib/src/async_expand.dart
natebosch: My thinking had been to avoid doing unnecessary work if there might not ever be any listeners, but I had been uncomfortable with the fact that whether the stream had listeners at at given time has an observable effect. I agree it was probably the wrong tradeoff to make. I'll see what this looks like if we keep listening on the outer stream.

natebosch: > I wouldn't delay subscribing to non-broadcast streams.

Why is that?

lrhn: If the returned stream is a broadcast stream, then it shouldn't matter if you cancel and re-listen, or stay subscribed and just throw away events for the same duration, it should give the same result (aka: the stream creates the same events at the same time, independently of whether there are listeners, it's purely an optimization to not call `add` while there is no listener.)
In that case, it's OK to not subscribe to inner broadcast streams until you need them, because it won't matter for which event subscribers see. You're using the broadcast property to unsubscribe until you need events again, instead of discarding them. For a properly implemented broadcast stream, that should be safe.

It does matter when you subscribe to an inner non-broadcast stream, because it changes the timing of events, so you should subscribe to it at the same time, aka. immediately, whether you have listeners or not. If you don't, the returned stream does not behave properly as a broadcast stream. It may not be obvious to listeners, because they don't know which events they should get, but we know! :)

lib/src/async_expand.dart
natebosch: I switched to `LinkedHashmap.identity`.

I looked for a pattern with a `Set` of some other class to hold the streams - but nothing felt cleaner to me. We currently use the Stream identity to clean up the subscriptions for streams that have ended, and it's convenient that we can do the same exact thing for this (assign into a Map) in the broadcast case regardless of whether this is a first or subsequent listen.

lib/src/async_expand.dart
natebosch: Done.

lib/src/async_expand.dart
natebosch: I agree, it's better that time of presence/absence of _other listeners_ on the stream do not impact the events that are seen. I updated the behavior so immediately listen on broadcast streams, and to avoid ever canceling the outer subscription. I updated the doc comment to warn about this behavior and added new tests.

lib/src/async_expand.dart
```diff
@@ -29,12 +30,18 @@ extension AsyncExpand<T> on Stream<T> {
   /// If the source stream is a broadcast stream, the result will be as well,
   /// regardless of the types of streams created by [convert]. In this case,
   /// some care should be taken:
-  /// -  If [convert] returns a single subscription stream it may be listened to
-  /// and never canceled.
-  /// -  For any period of time where there are no listeners on the result
-  /// stream, any sub streams from previously emitted events will be ignored,
-  /// regardless of whether they emit further events after a listener is added
-  /// back.
+  /// -  The source stream will be immediately listened to, regardless of
```
lrhn, line 33: This is worrisome behavior. Nothing should happen on a stream until someone listens to it, including the returned stream here.

lib/src/async_expand.dart
```diff
@@ -29,12 +30,18 @@ extension AsyncExpand<T> on Stream<T> {
...
+  /// never be canceled, even if all subscriptions on the result stream are
+  /// canceled.
+  /// -  If [convert] returns a single subscription stream while the result
+  /// stream has no listener it will be immediately listened to, and may be
+  /// drained without any events reaching a listener on the result stream. Only
+  /// events emitted by the inner stream _after_ a listen starts on the result
+  /// stream will be forwarded.
+  /// -  If [convert] returns a single subscription stream it will be listened
+  /// to and never canceled, even if all subscriptions on the result stream are
+  /// canceled.
```
lrhn, line 44: Auto-listening and never stopping worries me greatly.

Consider using and supporting multi-streams instead.

Every listen on the returned stream does a new listen on the source stream, and (if successful, so the source is not single-sub and used before) convert every event again and handle it like individual new streams.

If someone wants one of the expanded streams as a broadcast stream, they can call `asBroadcastsStream` on it, which gives you (something-like) the current behavior.

lib/src/async_expand.dart
```diff
@@ -29,12 +30,18 @@ extension AsyncExpand<T> on Stream<T> {
...
   ///
```
lrhn, line 45: That is, consider something like:
```dart
  Stream<S> concurrentAsyncExpand<S>(Stream<S> Function(T) convert) =>
      Stream.multi((controller) {
        Set<StreamSubscription> innerSubscriptions = {};
        var outerDone = false;
        var subscription = listen(
            (T event) {
              StreamSubscription<S> innerSubscription;
              try {
                var innerStream = convert(event);
                // If controller.addStream could be called multiple times,
                // just do controller.addStreamm(innerStream). Until then ...
                innerSubscription = innerStream.listen(controller.addSync,
                    onError: controller.addErrorSync);
              } catch (error, stack) {
                controller.addErrorSync(error, stack);
                return;
              }
              innerSubscriptions.add(innerSubscription);
              innerSubscription.onDone(() {
                innerSubscriptions.remove(innerSubscription);
                if (outerDone && innerSubscriptions.isEmpty) controller.close();
              });
            },
            onError: controller.addErrorSync,
            onDone: () {
              outerDone = true;
              if (innerSubscriptions.isEmpty) controller.closeSync();
            });
        controller
          ..onPause = () {
            subscription.pause();
            for (var innerSubscription in innerSubscriptions) {
              innerSubscription.pause();
            }
          }
          ..onResume = () {
            for (var innerSubscription in innerSubscriptions) {
              innerSubscription.resume();
            }
            subscription.resume();
          }
          ..onCancel = () => Future.wait([
                subscription.cancel(),
                for (var innerSubscription in innerSubscriptions)
                  innerSubscription.cancel()
              ])..then((_) => null);
      });
```

lrhn (commented):
In general, assuming that all streams are single-sub or broadcast isn't safe. Treating them differently is worrisome. Listening eagerly is very worrisome.

lib/src/async_expand.dart
natebosch: I made this change as a further step towards the principles discussed in https://github.com/dart-lang/stream_transform/pull/148#discussion_r895434468

The problem with not listening is that it becomes observable _when_ the first listener gets added on the stream. If we want the observable behavior that listener see on the stream to be independent of each other I think we have to listen on the outer stream immediately.

natebosch: Ah, I see the other comment about using a multi stream to preserve isolation between listeners... I like that idea.

I'll rework this.

lrhn: That too was a problem, caused by the returned stream being a broadcast stream.
Which might be the underlying problem that makes thing feel odd to me.

We don't have to listen to the outer stream immediately, if the individual listeners are completely separate computations. If the resulting stream is *never* a broadcast stream, then we don't have to worry about each listener seeing the same events at the same time, and it won't matter who listens first.

We can make the new stream a single-subscription stream. If you want to listen twice, wrap the original stream twice.
Or we can make it a multi-stream, where we treat each listen as a new wrapping of the original. 
Then each listener on the new stream sees a result computed from the events it would have seen if subscribing to the original stream at the same time. Because that's precisely what happens.
There is no interaction between them because they are not all looking at the same broadcast stream.



I guess the underlying problem is that I'm not sure what the user intent is for allowing multiple listeners on a stream.

A broadcast stream is fire-and-forget, listeners come and go. They are usually considered started before anyone listens, it's just that they know not to do anything. 

A single-subscription stream is intended for giving all its data to one listener, in order, starting when listened to.

It's not clear whether doing, say, `.throttle` is intended to *create* a broadcast or single-subscription stream.
It's definitely not obvious that it should always be the same as the source stream, because what you want depends on the use, not the source.
It's easier for something like `.map` where mapping the original events once or once per listener only affects how many times the `convert` function is called, not what the events are (assuming the `convert` function is side-effect free).

Any *history* based operation applied to a non-single-subscription stream can be treated in two ways:
* Use the global history and give everybody the same result.
* Use the individual history for each listener, only based on events emitted while it is listening.

The former either means listening immediately (or sooner?!) or having the result depend on when the first listener happens. The former is potentially expensive and wasteful. The latter is just weird and arbitrary.

And it's not clear why that behavior should only apply to broadcast source streams. 
It's not clear that the source being a broadcast stream is important, after all, you only listen to it once anyway, so why not do the same thing for all streams?

I guess it's just a weird model to apply a history-based transformation to something that *might* be running before you start. You should always assume that a stream does nothing until you listen, and you should (preferably, if possible) not listen until you know events are needed.

The best model for that is to not create a broadcast stream.

Creating a single-subscription stream is always valid. You can only use it once, but if the source is not a single-subscription stream, you can always create another. 
It's safe and understandable.

(And creating a multi-stream is just removing that one extra step, and allowing you to treat each `listen` as a new listen on a separate stream).

natebosch: > It's definitely not obvious that it should always be the same as the source stream, because what you want depends on the use, not the source.

I think the choice to maintain the broadcast status of the source stream was fairly arbitrary. I wonder if it would make sense to overhaul all the APIs to return single subscription streams consistently, and force callers to use `asBroadcastStream` if they want to get back to a broadcast stream.

lib/src/async_expand.dart
natebosch: > convert every event again and handle it like individual new streams

I was hoping to avoid that. Running a conversion per-event, instead of per-event-per-listener, is an intentional design. I find it strange that `map` is per-listener on broadcast streams.

lib/src/async_expand.dart
lrhn: I think there are APIs where it makes more sense to preserve broadcast-ness, like `map` or `where`, where (assuming pure functions), each listener will see the same thing, because the result depends entirely on the individual events emitted while you listen.

Synchronous per-event conversions, basically (to zero (`where`), one (`map`) or multiple (`expand`) new events, but all delivered immediately), where everything happens immediately when the original event is emitted, with no history or delay.

Even `asyncExpand` on `Stream` is not one of those, and I think it currently preserves broadcast-ness.

I'd consider preserving broadcast-ness for those, but it's also fair to use a single-sub stream and ask you to convert the stream twice which duplicates the computation for each listener. (That was actually a design goal for Dart 2.0, which slipped.) 

Or use a multi-stream, which acts like you did wrap it twice. 
Then I'd also use a multi-streams on all non-broadcast streams, in case the source is itself a multi-stream, so listening twice will work.

Always using a single-sub stream or a multi-stream is the simplest approach. Preserving broadcast-ness is really just an optimization.

lib/src/async_expand.dart
lrhn: I didn't even know that `map` was per listener. I guess we managed to get that one migrated for Dart 2.
(I hope it means that no conversion is happening if you pause a `map` of a broadcast stream. The stream keeps producing events, but they should be buffered at some level. Preferably *before* being converted, which would be one of the reasons to do it that way.)

mosuem:
@natebosch is this still WIP?

Closing this for now as it is stale. Feel free to reopen in case you still want to land it!

PR #151, opened 2022-06-14
Call out missed events in startWith Doc

natebosch:
The listen will not happen until the first event is delivered.


lib/src/concatenate.dart
```diff
@@ -82,6 +82,8 @@ extension Concatenate<T> on Stream<T> {
   /// Emits [initial] before any values or errors from the this stream.
   ///
   /// If this stream is a broadcast stream the result will be as well.
+  /// If this stream is a broadcast stream events which are emitted before
```
lrhn: Comma before "events".

Or consider rephrasing, it's a long sentence with the lede buried at the end.
Maybe:
> /// If this stream is a broadcast stream, the returned stream will only contain events
> /// of this stream that are emitted *after* the [initial] value has been emitted on
> /// the returned stream.

Technically, after the *done* event has been emitted by the `Future.asStream` stream, so "at some unspecified point after `initial` has been emitted". (In practice it's right after).

Consider just redoing it entirely as:
```dart
Stream<T> startWith(T initial) => Stream.multi((controller) { 
   controller..add(initial)..addStream(this).whenComplete(controller.close);
});
```
Then events from `this` happening between listen and emitting `initial` are not dropped, only delayed (will be buffered until `initial` has been delivered).

(Argh, just noticed that `Stream.multi`'s `.addStream` is not synchronous. That's a mistake. All stream controllers' `addStream` should emit stream events without further delay. Some stream refactoring needed for that, but I want to do it.)

macjohnny (approved):
LGTM

lib/src/concatenate.dart
```diff
@@ -82,6 +82,8 @@ extension Concatenate<T> on Stream<T> {
...
+  /// [initial] can be emitted asynchronously will be missed.
   Stream<T> startWith(T initial) =>
       startWithStream(Future.value(initial).asStream());
```
lrhn, line 89: (BTW, you can use `Stream<T>.value(initial)` instead of `Future<T>.value(initial).asStream()`. No promises that it has the precise same timing, but it *probably* does.)

kevmoo:
Any plans here, folks?

lrhn (approved):

lib/src/concatenate.dart
natebosch: I used your suggested phrasing.

lib/src/concatenate.dart
natebosch: Done.

natebosch: It does end up changing timing in a way the tests are sensitive too - it looks like it makes it miss even more elements. I'd guess we used to deliver in a microtask, and now deliver in an event. I'll keep this implementation as is for now.

PR #156, opened 2022-10-13
Add whereNotNull

natebosch:
This is useful during null safety migrations for the same reason as `Iterable.whereNotNull` from `package:collection`.


lrhn (approved):
Neat!

PR #159, opened 2022-12-12
Fix test compatibility with Dart 3

natebosch:
Move opt out test to separate directory so they will not run by default.

Collapse the chrome and VM tests into a single workflow step.
Add a separate workflow step to run the chrome and VM opt out tests for
only the oldest supported SDK.


.github/workflows/test-package.yml
```diff
@@ -56,9 +56,9 @@ jobs:
...
-        run: dart test --platform vm --test-randomize-ordering-seed=random
-        if: always() && steps.install.outcome == 'success'
-      - name: Run Chrome tests
-        run: dart test --platform chrome --test-randomize-ordering-seed=random
+      - name: Run tests
+        run: dart test -p chrome,vm --test-randomize-ordering-seed=random
         if: always() && steps.install.outcome == 'success'
+      - name: Run Legacy tests
+        run: dart test -p chrome,vm --test-randomize-ordering-seed=random legacy_test
+        if: always() && matrix.sdk == '2.14.0' && steps.install.outcome == 'success'
```
devoncarew, line 65: Perhaps also add a comment, or or around line 50, that when updating the 2.14 value there to also update here?

devoncarew (approved):

.github/workflows/test-package.yml
natebosch: Done.

PR #165, opened 2023-02-16
Forward errors from trigger in takeUntil

natebosch:
It doesn't make sense to leave the error unhandled in all cases. When we
can handle the error through the stream, do so, otherwise ignore it.

This also makes `takeUntil` more useful for cases like timeouts.


lib/src/take_until.dart
```diff
@@ -13,6 +13,12 @@ extension TakeUntil<T> on Stream<T> {
   /// which are emitted before the trigger, but have further asynchronous delays
   /// in transformations following the takeUtil, will still go through.
   /// Cancelling a subscription immediately stops values.
+  ///
+  /// If [trigger] completes as an error, the error will be forwared through the
```
lrhn: forwared -> forwarded

lrhn (approved):

PR #166, opened 2023-02-17
Use fake_async over waiting for timers with delays

natebosch:
On the web the timers may have some variance.

Replace all usages of `waitForTimer` utility to use `fakeAsync`.
Separate out the call to `Stream.listen` from `setUp` so it
happens in the fake async zone.

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -8,5 +8,6 @@ environment:
 
 dev_dependencies:
   async: ^2.5.0
+  fake_async: ^1.3.0
   lints: ^2.0.0
   test: ^1.16.0
--- a/test/audit_test.dart
+++ b/test/audit_test.dart
@@ -4,6 +4,7 @@
 
 import 'dart:async';
 
+import 'package:fake_async/fake_async.dart';
 import 'package:stream_transform/stream_transform.dart';
 import 'package:test/test.dart';
 
@@ -21,7 +22,7 @@ void main() {
       late StreamSubscription<int> subscription;
 
       group('audit', () {
-        setUp(() async {
+        setUp(() {
           valuesCanceled = false;
           values = createController(streamType)
             ..onCancel = () {
@@ -31,18 +32,23 @@ void main() {
           errors = [];
           isDone = false;
           transformed = values.stream.audit(const Duration(milliseconds: 6));
+        });
+
+        void listen() {
           subscription = transformed
               .listen(emittedValues.add, onError: errors.add, onDone: () {
             isDone = true;
           });
-        });
+        }
 
         test('cancels values', () async {
+          listen();
           await subscription.cancel();
           expect(valuesCanceled, true);
         });
 
         test('swallows values that come faster than duration', () async {
+          listen();
           values
             ..add(1)
             ..add(2);
@@ -52,6 +58,7 @@ void main() {
         });
 
         test('outputs multiple values spaced further than duration', () async {
+          listen();
           values.add(1);
           await waitForTimer(5);
           values.add(2);
@@ -60,6 +67,7 @@ void main() {
         });
 
         test('waits for pending value to close', () async {
+          listen();
           values.add(1);
           await values.close();
           expect(isDone, false);
@@ -68,6 +76,7 @@ void main() {
         });
 
         test('closes output if there are no pending values', () async {
+          listen();
           values.add(1);
           await waitForTimer(5);
           values.add(2);
@@ -79,17 +88,21 @@ void main() {
 
         test('does not starve output if many values come closer than duration',
             () async {
-          values.add(1);
-          await Future.delayed(const Duration(milliseconds: 4));
-          values.add(2);
-          await Future.delayed(const Duration(milliseconds: 4));
-          values.add(3);
-          await waitForTimer(6);
-          expect(emittedValues, [2, 3]);
-        }, onPlatform: const {'browser': Skip('Timer timing is unreliable')});
+          fakeAsync((async) {
+            listen();
+            values.add(1);
+            async.elapse(const Duration(milliseconds: 3));
+            values.add(2);
+            async.elapse(const Duration(milliseconds: 3));
+            values.add(3);
+            async.elapse(const Duration(milliseconds: 7));
+            expect(emittedValues, [2, 3]);
+          });
+        });
 
         if (streamType == 'broadcast') {
           test('multiple listeners all get values', () async {
+            listen();
             var otherValues = [];
             transformed.listen(otherValues.add);
             values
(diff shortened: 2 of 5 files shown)


natebosch:
We had problems with this test in the past and bumped the timing to account for it. #154 

@sigmundch - is it expected that a `Timer` started on the web may fire significantly late, with hardly any Dart level work happening during the delay?

sigmundch:
It's possible, but depends what you consider "significant". We compile things down to JavaScript's setTimeout, which can introduce delays for various reasons: https://developer.mozilla.org/en-US/docs/Web/API/setTimeout#reasons_for_delays_longer_than_specified

I'd be surprised of significant delays in the order of seconds, but not surprised on a few ms. A delay between 6ms and 10ms seems within many possible scenarios:
* precision errors - setTimeout is not exact and I've seen experiments for 1s sometimes finishing ±15ms, sometimes more. I don't know if that's because they accept a margin of error on large delays or if that's because the underlying system is not precise enough.
* weird internals of setTimeout - the link above, for example, talks about a 4ms delay when many nested setTimeouts are executed.

natebosch:
> * setTimeout is not exact and I've seen experiments for 1s sometimes finishing ±15ms, sometimes more.

This might be the explanation. I'm not sure how flaky the tests are currently, but I don't think it's worth extending the delays long enough that 15ms of variance doesn't cause flakes. I'll use `fake_async` instead.

pubspec.yaml
```diff
@@ -8,5 +8,6 @@ environment:
 
 dev_dependencies:
   async: ^2.5.0
+  fake_async: ^1.3.0
```
lrhn, line 11: I don't really *trust* fake_async, but for simple things, it's probably fine.
(Have seen too many people misuse it it internal tests, because of zones.)

test/audit_test.dart
```diff
@@ -31,73 +32,97 @@ void main() {
...
-        test('swallows values that come faster than duration', () async {
-          values
-            ..add(1)
-            ..add(2);
-          await values.close();
-          await waitForTimer(5);
-          expect(emittedValues, [2]);
+        test('swallows values that come faster than duration', () {
+          fakeAsync((async) {
+            listen();
```
lrhn, line 52: So nothing in `setUp` can be asynchronous (which is probably reasonable).

Maybe document that somewhere.

test/audit_test.dart
```diff
@@ -31,73 +32,97 @@ void main() {
...
+            values
+              ..add(1)
+              ..add(2)
+              ..close();
+            async.elapse(const Duration(milliseconds: 6));
```
lrhn, line 57: That feels like it's cutting it very close. I assume fake_async ensures that 6 ms is never +/- some microseconds, and is entirely deterministic, so if this works, it always works. (Right?)

test/audit_test.dart
```diff
@@ -31,73 +32,97 @@ void main() {
...
+        test('closes output if there are no pending values', () {
+          fakeAsync((async) {
+            listen();
+            values.add(1);
+            async.elapse(const Duration(milliseconds: 6));
+            values
+              ..add(2)
+              ..close();
+            expect(isDone, false);
+            async.elapse(const Duration(milliseconds: 6));
```
lrhn, line 95: Check `emittedValues` before and after elapse?

test/audit_test.dart
```diff
@@ -31,73 +32,97 @@ void main() {
...
-          expect(emittedValues, [2, 3]);
+            () {
+          fakeAsync((async) {
+            listen();
+            values.add(1);
+            async.elapse(const Duration(milliseconds: 3));
+            values.add(2);
+            async.elapse(const Duration(milliseconds: 3));
+            values.add(3);
+            async.elapse(const Duration(milliseconds: 7));
```
lrhn: Why 7 here, and not 6 like everywhere else?

test/audit_test.dart
```diff
@@ -31,73 +32,97 @@ void main() {
...
+              listen();
+              var otherValues = [];
+              transformed.listen(otherValues.add);
+              values
+                ..add(1)
+                ..add(2);
+              async.elapse(const Duration(milliseconds: 6));
+              expect(emittedValues, [2]);
+              expect(otherValues, [2]);
+            });
```
lrhn, line 134: Try doing the second listen after 3 mss, then add a second value there, and again after 3/4 more mss.
```dart
  listen()
  var otherValues = [];
  values.add(1);
  async.elapse(const Duration(milliseconds: 3));
  values.add(2);
  transformed.listen(otherValues.add);
  values.add(3);
  async.elapse(const Duration(milliseconds: 3));
  values.add(4);
  async.elapse(const Duration(milliseconds: 3));
  values..add(5)..close();
  async.elapse(const Duration(milliseconds: 6));
  // Check emitted and other values here
```
Should the two have the same events, starting from the first event?
(I guess they should, looking at the implementation. Consider whether `audit` on a broadcast stream should be a multi-stream instead of a single/broadcast stream, so each listener can be independent, and not depend on when other listeners did their listen.)

lrhn (approved):
Change LGTM.

Consider whether the audit/debounce/throttle transformations should be based on `Stream.multi` instead of single-sub/broadcasts streams.
These transformations are timer-based event filters on an event stream, and it seems odd that different filterings on the same source interfere with each other (the first one sets the timing, the others do not, so for example, an `audit` listen can get the first event before 6 ms has passed, possibly to its surprise.

Then it would also work with other multi-streams, which it currently doesn't (it only allows listening once to an `audit` on a multi-stream).

test/audit_test.dart
natebosch: Stuff in `setUp` can be async, but the stream listen can't happen there.

test/audit_test.dart
natebosch: Yes, `fake_async` is deterministic, this should always work.

natebosch:
> Consider whether the audit/debounce/throttle transformations should be based on `Stream.multi` instead of single-sub/broadcasts streams.

We should definitely consider this - we'll keep the discussion going in https://github.com/dart-lang/stream_transform/pull/148 or break out an issue for it.

pubspec.yaml
natebosch: I think our choices are either do this, or bump up the delays so that we can tolerate 15ms of error.

I do think this is the right type of use case for `fake_async`.

test/audit_test.dart
natebosch: Done.

test/audit_test.dart
natebosch: I probably hit `ctrl-a` with my cursor on the line and didn't notice...

Back to 6

test/audit_test.dart
natebosch: Done.

PR #171, opened 2023-05-23
Add a optional parameter 'waitAll' on combineLatestAll

tpouriel:
Allow combineLastestAll output stream to emit events even if all the input streams don't have emitted yet.

### DIFF
--- a/lib/src/combine_latest.dart
+++ b/lib/src/combine_latest.dart
@@ -174,7 +174,7 @@ extension CombineLatest<T> on Stream<T> {
   /// If the source stream is a broadcast stream, the result stream will be as
   /// well, regardless of the types of [others]. If a single subscription stream
   /// is combined with a broadcast source stream, it may never be canceled.
-  Stream<List<T>> combineLatestAll(Iterable<Stream<T>> others) {
+  Stream<List<T>> combineLatestAll(Iterable<Stream<T>> others, {bool waitAll = true}) {
     final controller = isBroadcast
         ? StreamController<List<T>>.broadcast(sync: true)
         : StreamController<List<T>>(sync: true);
@@ -193,8 +193,8 @@ extension CombineLatest<T> on Stream<T> {
       void handleData(int index, T data) {
         latestData[index] = data;
         hasEmitted.add(index);
-        if (hasEmitted.length == allStreams.length) {
-          controller.add(List.from(latestData));
+        if (hasEmitted.length == allStreams.length || !waitAll) {
+          controller.add(List.from(latestData.where((data) => data != null)));
         }
       }
 
--- a/test/combine_latest_all_test.dart
+++ b/test/combine_latest_all_test.dart
@@ -49,6 +49,44 @@ void main() {
       await first.close();
     });
 
+    test('emits latest values without waiting', () async {
+      final first = StreamController<String>();
+      final second = StreamController<String>();
+      final third = StreamController<String>();
+      final combined = first.stream.combineLatestAll(
+          [second.stream, third.stream], waitAll: false).map((data) => data.join());
+
+      // first:    a------b--------------------c--------d---|
+      // second:   ---1------------2-----------------|
+      // third:    -----------&----------%----|
+      // combined: a--a1--b1--b1&--b2&---b2%---c2%------d2%-|
+
+      expect(combined,
+          emitsInOrder(['a', 'a1', 'b1', 'b1&', 'b2&', 'b2%', 'c2%', 'd2%', emitsDone]));
+
+      first.add('a');
+      await tick();
+      second.add('1');
+      await tick();
+      first.add('b');
+      await tick();
+      third.add('&');
+      await tick();
+      second.add('2');
+      await tick();
+      third.add('%');
+      await tick();
+      await third.close();
+      await tick();
+      first.add('c');
+      await tick();
+      await second.close();
+      await tick();
+      first.add('d');
+      await tick();
+      await first.close();
+    });
+
     test('ends if a Stream closes without ever emitting a value', () async {
       final first = StreamController<String>();
       final second = StreamController<String>();


tpouriel:
I can provide more tests, docs and bump version if this base modification looks good to you...

natebosch:
Can you describe your use case?

lib/src/combine_latest.dart
```diff
@@ -193,8 +193,8 @@ extension CombineLatest<T> on Stream<T> {
       void handleData(int index, T data) {
         latestData[index] = data;
         hasEmitted.add(index);
-        if (hasEmitted.length == allStreams.length) {
-          controller.add(List.from(latestData));
+        if (hasEmitted.length == allStreams.length || !waitAll) {
+          controller.add(List.from(latestData.where((data) => data != null)));
```
natebosch, line 197: If `T` is a nullable type this won't do what we want.

If we want to stop supporting null unsafe code we might be able to use `whereType<T>()` here. If we wanted to keep supporting null unsafe code we'd need some more complexity. I think it's probably fine to bump our SDK constraint and drop support for null unsafe code.

Mike278: This would still be an issue in null safe code wouldn't it? When dealing with generics I thought the only way to unambiguously use `null` as a sentinel would be to put an `Object` bound on `T` (thus requiring callers to pick a different sentinel, which isn't ideal).

tpouriel:
My usecase is the following : 
I have a source stream emitting events describing the state of a model in a database.
The stream is consumed by a consumer, typically a Widget.
In between I have a some "middlewares" that inject more stuff into the model, but some of the middleware can be quite slow if they have a lot of computation.

Here is what it looks like :

```
class Model {
...
}

interface Producer {
   Stream<List<Model>> streamFromDb();
}

interface Middleware {
   Stream<List<Model>> applyModification(Stream<List<Model>> source);
}

class StreamManager {
   final List<Middleware> middlewares = [...];
   final Producer producer;
   
   Stream<List<Model>> getAugmentedStream() {
      final source = producer.streamFromDb().asBroadcastStream();
     return middlewares.first.applyModification(source).combineLatestAll(middlewares.subList(1).map((middleware) => middleware.applyModification(source)).map(_resolveModifications);
   }
}

interface Consumer  {
   // getAugmentedStream is called here
}

```

lib/src/combine_latest.dart
tpouriel: Decision is your's but I think null unsafe code is quite outdated now

natebosch: > This would still be an issue in null safe code wouldn't it?

Yes, that's a good point - it would make it look like some false `null` were added in the source streams when the type happens to be nullable.

We probably need to use `hasEmitted`

```dart
        if (hasEmitted.length == allStreams.length) {
          controller.add(List.from(latestData));
        } else if (!waitAll) {
          var emittingIndexes = hasEmitted.toList()..sort();
          controller
              .add([for (var index in emittingIndexes) latestData[index] as T]);
        }
```

lib/src/combine_latest.dart
```diff
@@ -174,7 +174,7 @@ extension CombineLatest<T> on Stream<T> {
   /// If the source stream is a broadcast stream, the result stream will be as
   /// well, regardless of the types of [others]. If a single subscription stream
   /// is combined with a broadcast source stream, it may never be canceled.
-  Stream<List<T>> combineLatestAll(Iterable<Stream<T>> others) {
+  Stream<List<T>> combineLatestAll(Iterable<Stream<T>> others, {bool waitAll = true}) {
```
natebosch, line 177: We will need to update the Doc comment and add a CHANGELOG entry

natebosch: @lrhn - any thoughts on the argument name? If I was writing the PR I'd probably go with `emitPartial`.

This is likely a rarely used argument, so a long name should be fine.

lrhn: The default value of an optional boolean should be `false`, if possible, so the name should match that. 
The `emitPartial` can work. I think I prefer `allowPartial`, because the `emit` needs to be explained.)
If that's what it means &mdash; the parameter should be documented.

But I don't like the design to begin with, so I'd go a different way than a boolean flag.

Removing the `null` values removes the correspondence between value and source position.
In the complete lists, values are in a specific predictable position.
In the partial lists, they're not, the positions might as well be random.

I'd rather have a separate function which emits partial *and* complete lists, with return type `Stream<List<T?>>`.

Heck, even a function which which has two streams, with return type `(Stream<List<T?>>, Stream<List<T>>)`, and then you can listen to whichever you want (the first one is always broadcast, so we're sure we get to the second one without anyone being able to pause first).

(But I'd also make them both extension methods on `Iterable<Stream<T>>`, not on `Stream<T>` taking more streams as argument, because the first stream isn't special. I think I might even have code for it somewhere, or maybe I'm confusing it with "latest" on records of streams.)

natebosch: > But I don't like the design to begin with, so I'd go a different way than a boolean flag.
> 
> Removing the `null` values removes the correspondence between value and source position.

I think this is a caveat that we could document - it only impacts the partial lists emitted before all inputs have emitted a value.

lrhn: The only way to know if a list is partial is knowing the number of streams in the original iterable.

I just don't think it's a good design to mix the partial and complete results in the same function. The two exposed behaviors are too different, even if the events are found by the same underlying logic.

So: One function for complete lists, non -nullable element types, one or two which also include partial lists, one can be nullable, the other can remove the nulls and not preserve position, but the latter is easily computer from the former, so if probably go with only one.

It doesn't have to be the same function. Boolean flags which change behavior, in this case straight away from the original design, may be better off as a separate function.

mosuem:
Closing as the [dart-lang/stream_transform](https://github.com/dart-lang/stream_transform) repository is merged into the [dart-lang/tools](https://github.com/dart-lang/tools) monorepo. Please re-open this PR there!

PR #187, opened 2024-06-12
Remove default onData for transformByHandlers

natebosch:
Towards dart-lang/tools#1771

The default callback is not used in practice outside of tests, but it
included a cast to satisfy the type signature which looks worrying for
this type of library.

Make the callback required and add explicit callbacks (without a cast
since the types are consistent) in the tests which used it.


lrhn (approved):
Good fix.

(I'll have more comments, but this one was the first I noticed.)

PR #188, opened 2024-06-13
Replace Future.wait with Iterable<Future>.wait

natebosch:
Towards dart-lang/tools#1771

### DIFF
--- a/lib/src/aggregate_sample.dart
+++ b/lib/src/aggregate_sample.dart
@@ -138,7 +138,7 @@ extension AggregateSample<T> on Stream<T> {
         // Handle opt-out nulls
         cancels.removeWhere((Object? f) => f == null);
         if (cancels.isEmpty) return null;
-        return Future.wait(cancels).then((_) => null);
+        return cancels.wait.then((_) => null);
       };
     };
     return controller.stream;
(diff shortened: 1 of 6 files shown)


kevmoo (approved):

lib/src/aggregate_sample.dart
```diff
@@ -138,7 +138,7 @@ extension AggregateSample<T> on Stream<T> {
         // Handle opt-out nulls
         cancels.removeWhere((Object? f) => f == null);
         if (cancels.isEmpty) return null;
-        return Future.wait(cancels).then((_) => null);
+        return cancels.wait.then((_) => null);
```
lrhn, line 141: Could all these `(_) => null` use the same top-level helper function, `void voidCallback(void _) {}`?
That should ensure we only have one function instance, instead of one per call point.

lib/src/switch.dart
```diff
@@ -128,7 +128,7 @@ extension SwitchLatest<T> on Stream<Stream<T>> {
           // Handle opt-out nulls
           ..removeWhere((Object? f) => f == null);
         if (cancels.isEmpty) return null;
-        return Future.wait(cancels).then(_ignore);
+        return cancels.wait.then(_ignore);
```
lrhn, line 131: Can be `(_) => null` like the rest, or use the same helper function.

lrhn (approved):

